PR #3, opened 2016-03-24
added function voxelize()

ptompalski:
this is to rather test if I didn't screw anything up on git/github, as I'm doing this for the first time.
the function I added is a simple conversion of lidar point cloud into voxels, that I use from time to time. The next one I plan to add is based on this one.

Please take a look at the code and let me know if it is ok. 
If yes, I plan to add little more, maybe also a plot method for the voxel output.


Jean-Romain:
Thanks. I understand the algorithm. It's exactly how I would code the function. But I have some advices before to merge the pull request.

`if (class(lidar) != "Lidar") stop("Input data not of class Lidar")`

This is not really usefull. It's better to use the R dispactcher both on S3 or S4 functions. I can modify that myself if you want.

     mround <- function(x,base=10){
          base*round(x/base)
      }

This function already exists and is intensively used in the package. It comes from plyr `plyr::round_any` you do not need to add a new undocumented function. But this function is for vector only. Your code become (using piping operators from magrittr package)

    voxels <- pointcloud@data[, 1:3, with = FALSE]

    voxels$X %<>% round_any(voxel_size)
    voxels$Y %<>% round_any(voxel_size)
    voxels$Z %<>% round_any(voxel_size)

Furthermore it would enable to have different voxel resolutions on X Y Z (Why not. When I was working on X rays tomography, the voxels resolution in Z was not the same than in X Y).

` # this check below can be probably removed`

Yes, if the lidar object does not have X Y and Z col names it means that you edited it by hand. This object will not be a valid object anymore and every functions of the package will fail.

And please force the function `unique`to be the one from package `data.table`. Data tables are much more efficient than R base. Here the function is maybe the one from R base.

Please make the modification. For the R dispatcher I can create a new S4 function for lidar object myself if you want. But if you want to add a function for plotting voxels or to process a table of voxel it could be good to looks how R dispatcher works.

I'm waiting for knowing what do you do with your voxels :-)

JR

ps: For merging pull request I have a button `merge pull request` it looks very easy.

ptompalski:
Thank Jean-Romain, I'll do all the corrections. 
However, I couldn't find anything about the R dispatcher you mentioned, and this sounds new to me. Could you explain shortly or provide a link with some explanation?

I am glad that you liked the general approach:)

Jean-Romain:
When you use a function R automatically detect the class of the data you want to process and call the appropriated function for this kind of object (if you wrote the proper code to do that). The best example is `plot`. `plot` is a generic function which do nothing but which call the proper `plot` function depending on what you want to plot.

- `plot(x,y)` with x and y two vectors call `plot.xy` (I assume)
- `plot(x)` with x a data.frame call `plot.data.frame`
- `plot(x)` with x a linear model object call  `plot.lm`
- `plot(x)` with x a lidar object call  `plot.lidar`

You do not need to test the class of the object in each specific fucntion. The class have been tested by the R dispatcher.

The S3 way to do that is using `UseMethod`.You define a generic function and the other function inherit from this function usind an extented name. Look at an easy example on `round_any` https://github.com/hadley/plyr/blob/master/R/round-any.r

    x = 1:3
    class(x) = "unkwnon"
    plyr::round_any(x)

> Error in UseMethod("round_any") : 
  pas de méthode pour 'round_any' applicable pour un objet de classe "unkwon"

- [UseMethod](https://stat.ethz.ch/R-manual/R-devel/library/base/html/UseMethod.html)
- [HadleyWikckam's blog about R dispacth](http://adv-r.had.co.nz/S3.html)

The S4 way is different and look more like the C++ way using function signatures. But the idea is the same. You create a generic S4 function using `setGeneric()`. All my generic functions are grouped in the file `setGeneric.r`. Then, you define the function designed for the type of data you want to process. Here the example of `gridMetric`

    setGeneric("gridMetrics", function(obj, res, func, start=c(0,0), option = NULL){standardGeneric("gridMetrics")})

    setMethod("gridMetrics", "Lidar",
	    function(obj, res, func, start = c(0,0), option = NULL)
	    {
                 .....

-  [HadleyWikckam's blog about S4](http://adv-r.had.co.nz/S4.html)

In both cases if you input an object for which there is no specific function you will receive an error. But you do no need to manage this error. It's the R's job.

ptompalski:
Hi Jean-Romain,

I edited the function code and this is how it looks like now (see below). I thought I show you the code before asking for another pull request, so you can check if the changes are ok.
I didn't know if I should add the S3 method only, or S3 and S4. For now it is S3 only. Is this ok? 

I added option to choose different voxel size for each dimension, with support for single value. 


```
voxelize <- function(pointcloud,voxel_size = 1) {
  UseMethod("voxelize")
}

voxelize.Lidar <- function(pointcloud,voxel_size = 1) {
  
  #check input voxel_size param
  #must be length= 1 (all sides equal) or length=3 (different dimensions of each side of the voxel)
  if (!(length(voxel_size) %in% c(1,3))) stop("voxel_size must be a single value or a vector of length 3")
  
  if(length(voxel_size == 1)) voxel_size <- rep(voxel_size,3)
  
  #keep only XYZ
  voxels <- pointcloud@data[, 1:3, with = FALSE]
  
  # round to nearest vox size
  voxels$X %<>% plyr::round_any(voxel_size[1])
  voxels$Y %<>% plyr::round_any(voxel_size[2])
  voxels$Z %<>% plyr::round_any(voxel_size[3])
  
  #remove duplicates
  voxels <- unique(voxels) 
  return(voxels)
}

```

Jean-Romain:
Hi, 

I like it. I have two tiny request. 

Please change `pointcloud` to `lidar` or to `object`or anything which is not confusing. The function expect a `Lidar` object not a matrix or a data.frame of points. Using `pointcloud`could be confusing for user which read the doc too quickly. I like semantic code which is understandable itself :smile: 

    if (!(length(voxel_size) %in% c(1,3)))

    if (length(voxel_size) > 3)

This is simpler isn't it? (And incredibly faster but here with 3 or 4 figures we don't care)

> I didn't know if I should add the S3 method only, or S3 and S4. For now it is S3 only. Is this ok? 

Me neither. I use S4 for every function related to Lidar objects. But S3 works too and I don't know exactly the difference. It does not change anything for the user. So it's ok for me. If I learn more about S3/4 maybe I will change this function, or mine.

The function is ok for me. But if you want you can add an interesting feature for a next update in the future. In this function you keep only spatial coordinates. You loosed every other informations. It's a pity. You can aggregate the data to attribute, for each voxels, a statistic of intensity, scan angle etc.  Mean intensity, mean scan angle etc. voxelize could be a 3D version of `gridMetrics`.

ptompalski:
Hi Jean-Romain
this look awesome!
Did you already implement that? If yes my input is no longer required here, right?

Jean-Romain:
I didn't do it but I will. I'm currently too lazy and the case without aggregation (return only X Y Z) is surprisingly not the easiest to manage. Maybe I will do that this week-end

Jean-Romain:
I added a function `voxelize` which work like `gridMetrics`. Documentation has been copied/pasted from `gridMetrics`with some adaptations. I didn't really check what I wrote :anguished: . `voxelize` returns a `data.table` with the class `voxels` and it can be plotted easily like a `Lidar` object. The plot function has a parameter `display` enabling to plot nice shaded cube like in my previous post. This plot method is **extremely slow** and cannot be used for a large number of voxels.

I didn't manage the possibility to have different resolution on X Y Z. It create more problem than the usefulness of this feature. I cannot allow to *only voxelize* the cloud of point. It is required to compute at least one metric

    LASfile <- system.file("extdata", "Megaplot.las", package="lidR")
    lidar = LoadLidar(LASfile)
    subset = lidar %>% clipCircle(685000, 5018000, 10)
    
    plot(subset)
    voxels = voxelize(subset, 1 , length(Z))
    
    plot(voxels)
    plot(voxels, display = "cubes")
    
    # here we can see that is it no very usefull to voxelize a cloud with a low density of points there is only one point in each voxel
    plot(voxels, color = "V1", display = "cubes")

PR #30, opened 2016-11-19
Fix: roi_query output names and order

floriandeboissieu:
roi_query was assigning wrong roi names to output and wasn't reordering the output to roinames before return. These modifications are fixing these issues.

### DIFF
--- a/R/roi_query.r
+++ b/R/roi_query.r
@@ -43,6 +43,7 @@
 #' @param r2 numeric or vector. A radius or a set of radii of plots. If r2
 #' is provided, the selection turns into a rectangular ROI. If r = r2 it is a square obviouly.
 #' @param roinames vector. A set of ROI names (the ID of the plots for example)
+#' @param cl a cluster object, created by package parallel or package snow. If NULL, no parallel computing (default).
 #' @param ... additionnal parameters for \link[lidR:readLAS]{readLAS}
 #' @return A list of LAS objects
 #' @export roi_query
@@ -64,11 +65,11 @@
 #' # Return a List of 30 square LAS objects of 50x50 m
 #' catalog %>% roi_query(X, Y, R, R)
 #' }
-setGeneric("roi_query", function(obj, x, y, r, r2 = NULL, roinames = NULL, ...){standardGeneric("roi_query")})
+setGeneric("roi_query", function(obj, x, y, r, r2 = NULL, roinames = NULL, cl=NULL, ...){standardGeneric("roi_query")})
 
 #' @rdname roi_query
 setMethod("roi_query", "Catalog",
-          function(obj, x, y, r, r2 = NULL, roinames = NULL, ...)
+          function(obj, x, y, r, r2 = NULL, roinames = NULL, cl=NULL, ...)
           {
             CIRCLE = 0
             RECTANGLE = 1
@@ -94,35 +95,64 @@ setMethod("roi_query", "Catalog",
             
             cat("Extracting data...\n")
             
-            p = dplyr::progress_estimated(nplot)
-            output=lapply(X=c(1:nqueries),function(i){
-              require(lidR)
-              query = lasindex[i]
-              
-              X     = query$X[[1]]
-              Y     = query$Y[[1]]
-              r     = query$r[[1]]
-              r2    = query$r2[[1]]
-              files = query$tiles[[1]]
-              
-              lidar = readLAS(files,...) #
-              output=vector("list", length(X))
-              
-              for(j in 1:length(X))
-              {
-                if(type == CIRCLE)
-                  output[[j]] = clipCircle(lidar, X[j], Y[j], r[j])
-                else
-                  output[[j]] = clipRectangle(lidar, X[j]-r[j], Y[j]-r2[j], X[j]+r[j], Y[j]+r2[j])
+            if(is.null(cl)){
+              p = dplyr::progress_estimated(nplot)
+              output=lapply(X=c(1:nqueries),function(i){
+                require(lidR)
+                query = lasindex[i]
                 
-                p$tick()$print()
-              }
-              names(output)=query$roinames[[1]]
-              rm(list="lidar")
-              gc()
-              return(output)
-            })
-            
+                X     = query$X[[1]]
+                Y     = query$Y[[1]]
+                r     = query$r[[1]]
+                r2    = query$r2[[1]]
+                files = query$tiles[[1]]
+                
+                lidar = readLAS(files,...) #
+                output=vector("list", length(X))
+                
+                for(j in 1:length(X))
+                {
+                  if(type == CIRCLE)
+                    output[[j]] = clipCircle(lidar, X[j], Y[j], r[j])
+                  else
+                    output[[j]] = clipRectangle(lidar, X[j]-r[j], Y[j]-r2[j], X[j]+r[j], Y[j]+r2[j])
+                  
+                  p$tick()$print()
+                }
+                names(output)=query$roinames[[1]]
+                rm(list="lidar")
+                gc()
+                return(output)
+              })
+            }else{
+              parallel::clusterExport(cl, varlist=c(lsf.str(envir = globalenv()), ls(envir = environment())), envir = environment())
+              output=parallel::parLapply(cl,X=c(1:nqueries),function(i){
+                require(lidR)
+                query = lasindex[i]
+                
+                X     = query$X[[1]]
+                Y     = query$Y[[1]]
+                r     = query$r[[1]]
+                r2    = query$r2[[1]]
+                files = query$tiles[[1]]
+                
+                lidar = readLAS(files,...) #
+                output=vector("list", length(X))
+                
+                for(j in 1:length(X))
+                {
+                  if(type == CIRCLE)
+                    output[[j]] = clipCircle(lidar, X[j], Y[j], r[j])
+                  else
+                    output[[j]] = clipRectangle(lidar, X[j]-r[j], Y[j]-r2[j], X[j]+r[j], Y[j]+r2[j])
+                }
+                names(output)=query$roinames[[1]]
+                rm(list="lidar")
+                gc()
+                return(output)
+              })
+            }
+
             output=unlist(output)
             ## set back to the original order
             output=output[match(roinames,names(output))]
--- a/R/roi_query.r
+++ b/R/roi_query.r
@@ -68,65 +68,67 @@ setGeneric("roi_query", function(obj, x, y, r, r2 = NULL, roinames = NULL, ...){
 
 #' @rdname roi_query
 setMethod("roi_query", "Catalog",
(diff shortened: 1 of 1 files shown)


Jean-Romain:
Hi,

2 commits but two very different features.

**output order**

You modifiied the indentation of each line so all the line are considered as modified. It difficult to see what changed. I understand your code but I can't see what was the problem and what is your solution. Reading the two codes they seem to do the same things for me. Could you explain me?

**parallel computation**

First: the single thread process was a real choice. Parallel process have a cost: memory. During the process each step of the loop can open 1 to 4 tiles (event more since you previous pull request). Parallelization on n cores implies that  4n files could be loaded in one time. If the user's computer is a regular computer with regular las file this is very dangerous.

Second: your code is not good. You copy/paste twice the same code. This is not admissible. Something must be factorize in an external function. Actually, parallelization on one core is a regular loop. So the single thread code in the same code than the parallel one... but with a single core cluster.

Third:  user must provide a cluster in input. I disagree with that. It means that the user must be comfortable with parallelization concepts. A lot of users are not comfortable with simple programming concepts. The parallelization must operate transparently. My opinion is that our job is to make things easy.

floriandeboissieu:
Hi,
sorry, I was aiming to propose 2 features in 2 commits so you can choose, but it doesn't seem to be possible. I'll redo the pull-requests separate.
# output order
Before code modification the output names were given at initialization of output, but as the queries are grouped by the tiles they are querying the input roinames order might not be the computation order. For example query 1 might be associated with query 3, thus the roinames must be assign inside the processing loop and eventually reordered at the end of the processing to correspond to the input roinames order.
# parallel computation
I agree with you on the 3 points, however:
1,3) I made the multi-core computation optional so that in the case it is not used it is still computed with a single thread. People who don't know well how to manage a cluster object could keep in this single threaded solution. To answer your concerns on memory, the memory necessary depends on the size of the tiles, the architecture of the computer and the what the user wishes to assign to the processing. That's why I think it as a user-defined parameter (e.g. you don't want to use all your cores if working on other things while processing). A simple example could help anyone to use that option (I can write it if you agree). Whatever the solution chosen, multi-core processing is very useful in that kind of task (e.g. I had to extract 160 ROI on a dataset, which took me less than 10 min. on 6 cores and would have taken me almost an hour on 1 core). 
2) I divided the code into 2 parts (single core, multi-core) because I thought useless to open a cluster of one core (taking time for nothing). I used 2 times the same code in a first place because I didn't managed to make it work well with a single function defined aside, but you are right I shouldn't have made a pull request until it is well coded. I'll try to manage that point soon.

Hoping it answers your questions.

Jean-Romain:
**ouput order**

Got it! You're right. I've never seen the bug because I don't use this code myself and I didn't run enough tests. I will read your code with such explanations in mind. But yes one pull request by feature. I can easily accept or reject the whole pull request but not only some part of the pull request.

**multithreading**

I agree with you about the optional multi threatening. If the default behavior is 1 core there is no problem. I skipped this argument because I expected you to freely provide some opposite arguments. But I'm basically ok with optional multithreading. But it must be done properly.

>  I didn't managed to make it work well with a single function defined aside,

Because you run on Windows. Windows cannot deal with forked processes. It makes everything more difficult. Basically you must export the package in each core in addition to global_env to provide the function to each process. I'm not comfortable with clustering. I usually use `mclapply` for my own codes which relies on forked process and make everything easier. But it does not work on Windows.

On solution is probably to export only the useful function in the cluster not the whole package. You can maybe get inspired by the function `process_parallel`. Both the doc and the code. But I'm not a specialist on parallelization.

Also I don't like the cluster input for two reasons:

1. the complexity as I said. A parameter `ncore` would be better in my opinion
2. it implies that linux users must be bored by clusters when they can rely on forked processes.

Jean-Romain:
>  because I thought useless to open a cluster of one core (taking time for nothing)

I don't know. Didn't benchmark it. I think we don't care. I will make tests.

floriandeboissieu:
Hi again,
I run the code under linux (ubuntu 12.04), and actually I used mcapply a while ago but the multithreading wasn't working well checking the core usage (1 or 2 cores used instead of the 6 asked). Seeing your code in process_parallel, expecting maybe some improvement since then, but still few cores are used. Moreover, I used some R code on a HPC, therefore I oriented my programming on cluster objects (platform independent and transferable easily to a cluster, with a job scheduler defining specifically memory, cores, timer,etc. that will be used). 
Here the problem has been more in the way of passing correctly the parameters to the lapply function. I'll try to make it clean with ncores if you prefer and send you a new pull request. Cheers,

Jean-Romain:
mclapply use my 4 cores. I don't experiment your problem. But as I said I'm not a specialist about parallelization. I only observe that parallelization is easy since I use forked processes.

Anyway in this case we do not care of the method because the user do not need to export anything itself. All can be done into the function. So cluster or fork is not important as long as everything is hidden to the user.

Jean-Romain:
> I don't know. Didn't benchmark it. I think we don't care. I will make tests.

Here my conclusions based on the following code. fork-exec processes win both in time computation and ease of use while clusters are a little bit slower and very hard to manipulate. Single core parallelization is almost the same as regular loop.

```r
n = 16

pause = function(x) { 
  t = 0.5+runif(1)
  Sys.sleep(t)
  return(t)
}

f = function() {
  lapply(1:n, pause)
}

g = function(core) {
  mclapply(1:n, pause, mc.cores = core)
}

h = function(core) {
  cl <- makeCluster(core)
  clusterExport(cl, varlist = NULL)
  out = parLapply(cl, 1:n, pause)
  stopCluster(cl)
  return(out)
}

bench = microbenchmark::microbenchmark(f(), g(1), h(1), g(4), h(4), times = 3)
```

```
Unit: seconds
 expr       min        lq      mean    median        uq       max neval
  f() 14.619520 15.436175 15.813373 16.252831 16.410300 16.567769     3
 g(1) 16.130614 16.359400 16.674677 16.588186 16.946708 17.305231     3
 h(1) 15.088411 15.277482 15.799586 15.466554 16.155174 16.843794     3
 g(4)  4.689136  4.790983  4.918875  4.892830  5.033745  5.174661     3
 h(4)  5.334808  5.486043  5.588740  5.637278  5.715707  5.794135     3
```

multithread

Jean-Romain:
I made a nice discovery. I can't explain that. I think it's called hyper-threading. I might be based on logical cores. My computer has 4 cores (intel core i3). I parallelized the previous code for n = 64 on 1, 4, 8, 16, 32 and 64 cores based on fork method. Look at the results! So I don't understand your point when you say that `mclapply` does not use all your cores. I can use more than all my core! 
You can do the same with SOCK cluster but after 4 cores there is no more improvement. It uses only the 4 physical cores.

multithread

floriandeboissieu:
In the past I used mclapply during while, but I experienced loosing some nodes or using fewer processors than asked, and I never found out why. When happening, I made those test between parLapply and mclapply, and my conclusion in those cases was always that parLapply was more reliable. However, I tested here with both configuration and opening/closing the cluster takes ~7s. Everything seems to work well (good number of nodes, etc.) similarly in both mclapply and parLapply. The results are below (sorry didn't made the graphs). mclapply is going a little faster considering you don't have to open a cluster. I can make the option function of the OS like it is in process_parallel.r.

```
## with parLapply, cluster opened outside roi_query
cores=6
cl = parallel::makeCluster(cores, outfile = "")
library(microbenchmark)
microbenchmark::microbenchmark(roi_query(obj=lascat,x = placettes$Xc,y = placettes$Yc,r=15,roinames = placettes$Id_plac,cl=cl),times=3)
parallel::stopCluster(cl)
# Unit: seconds
# expr      min       lq     mean   median
# roi_query(obj = lascat, x = placettes$Xc, y = placettes$Yc, r = 15,      roinames = placettes$Id_plac, cl = cl) 57.22322 57.44327 59.54857 57.66332
# uq      max neval
# 60.71125 63.75918     3

## with parLapply, cluster opened inside roi_query
cores=6
library(microbenchmark)
microbenchmark::microbenchmark(roi_query(obj=lascat,x = placettes$Xc,y = placettes$Yc,r=15,roinames = placettes$Id_plac,cl=cores),times=3)
# Unit: seconds
# expr      min      lq     mean   median
# roi_query(obj = lascat, x = placettes$Xc, y = placettes$Yc, r = 15,      roinames = placettes$Id_plac, cl = cores) 65.43987 65.7611 66.65712 66.08234
# uq      max neval
# 67.26575 68.44916     3

## with mclapply
cores=6
library(microbenchmark)
microbenchmark::microbenchmark(roi_query(obj=lascat,x = placettes$Xc,y = placettes$Yc,r=15,roinames = placettes$Id_plac,cl=cores),times=3)
# Unit: seconds
# expr      min      lq     mean   median
# roi_query(obj = lascat, x = placettes$Xc, y = placettes$Yc, r = 15,      roinames = placettes$Id_plac, cl = cores) 59.60916 60.4901 61.84714 61.37104
# uq      max neval
# 62.96614 64.56123     3

```

I also used hyper-threading during a while, but it may decrease performance when increasing the number of trheads, depending on the job to be done. There also I had bad experiences, so without the time to dig more into it I retrained myself to the number of available cores...

Jean-Romain:
In this case we don't care of the parallelization method because there is nothing to export into the cluster for the user. My reservations concern only the copy/pasted code. I tested it and I confirmed that parallelization on a single core is the regular loop. So we can only use the parallel code adding a input parameter `mc.core = 1`. I think it makes sense.

About the output order I'm still unable to see the bug reading the codes. I need to make tests to understand the problem. I trust you but I had to run the code to see the bug with my own eyes.

About hyperthreading I don't know anything. My question was only: *what happen if the input num of cores is higher than the actual number of cores?* (1) crashes (2) uses 4 cores. The awnser was the third one: it uses more core than those I have. I cannot understand. I asked a friend to explain me this behavior.

floriandeboissieu:
About the order of output, I made a test with several ROI on the same tile distributed among other ROI in other tiles, which has confirmed a wrong naming of the output. I made the pull-request separate for that update.
About the parallelization, all the variables and functions used in parLapply must be exported to the cluster. However, in my opinion the cross-platform capacity of parLapply makes it the good solution. I will start with that and up to you to merge it now and change it afterwards if you feel it is better to use mclapply.
To answer your hyper-threading question: a thread is a subset of a process (i.e. at the software level), and you can have any number of threads per process leading to multi/hyper-threading. It was already used with 1 core. When using a cluster object it generates multiple processes (multiprocessing) that are managed by a master process, it is at the OS level, and one core can only run one process at the same time. I hope it helps, I don't have any document at the moment that would explain it clearly but if I find one I'll send it to you.

Jean-Romain:
I agree. We don't care of the parallel method means we must use the crossplatform one. So `parLapply`. As long as it's easy for the end user it's good. I used both `mclapply` and `parLapply` in `process_parallel` to be crossplatform and easy to use for unix alike users (i.e. do not take care of variable export).

PR #31, opened 2016-11-19
method "[" for Class Catalog

floriandeboissieu:
Usefull to make a subset of class Catalog.


Jean-Romain:
Make sense. Actually S4 class is useless and a classical S3 class would be better but this is how it work currently.

PR #32, opened 2016-11-20
Fix: Wrong pulseID detection for a tiled of a multiple flightline

floriandeboissieu:
Solution: aggregate on gpstime, but increases tile loading time as NumericRounding must be set to 0 to have a correct comparison of gpstime. Maybe detect_pulse should not be compulsory but optional at LAS reading.


floriandeboissieu:
The case considered in this fix is an acquisition cut into tiles. Some echos of same pulse may belong to different tiles. For example, let's consider 2 successive pulse at the border of the tile, one inside, the second outside. Let's say both are leading to 2 echos. If the second echos of the pulse inside tile has been removed for any reason, the second echo of the second pulse with first echo in another tile may be wrongly associated to the first pulse if based only on ordering along gpstime and detecting increasing ReturnNumber as it was done. gpstime seems to be the only variable to identify pulse. Actually multiple flightline has nothing to do with the problem (wrongly stated in first message) as gpstimes are completely different between flightlines.

Jean-Romain:
Hi,

I don't understand your code. I didn't test it but it seems to attribute an ID to each point not to each pulse. Isn't it ? 

Furthermore you removed each code like `gpstime <- pulseID <- ReturnNumber <- NULL` in all your pull requests. These kind of lines are useful. It's a trick to make the R code validator happy when we use lazy evaluations. Look this code:

```
obj@data[,pulseID:=.GRP,by=gpstime]
```

When we run a check on this kind of code it generate a 2 notes about the undefined variable `pulseID` and `gpstime`. The internal behavior is a lazy evaluations but R cannot guess it. So R supposes these two variables are undefined. This is why they are defined at the begining of the function. To make R happy and pass every CRAN tests.

Jean-Romain:
You're right, some removed multiple returns (for any reason) could lead to misdetection in some cases. But I can't see how you can detect pulses based only on gpstime. It introduces much more misdetection that the first method.

Jean-Romain:
I got it. I didn't know that each pulse have exactly the same timestamp. I believed that each point have a different timestamp even those coming from the same ray. My mistake. So your code make a lot of sense.

It might work at least in the dataset provided in the package.  But are you sure it always true? I mean there is no information in the las specification about this subject:

> GPS Time: The GPS Time is the double floating point time tag value at which the point was acquired. It is GPS Week Time if the Global Encoding low bit is clear and Adjusted Standard GPS Time if the Global Encoding low bit is set (see Global Encoding in the Public Header Block description)

Please confirm this last point and redo the removed line such as `gpstime <- pulseID <- ReturnNumber <- NULL` and ok good job.

R/indentify.r
```diff
@@ -120,14 +111,12 @@ setMethod("detect_flightline", "LAS",
 #' @return Return nothing. The original object is modified in place by reference.
 #' @export detect_scanline
 #' @importFrom data.table setorder
-setGeneric("detect_scanline", function(obj){standardGeneric("detect_scanline")})
+setGeneric("detect_scanline", function(obj, dt = 30){standardGeneric("detect_scanline")})
```
Jean-Romain: No need of dt in this function

R/indentify.r
```diff
@@ -38,29 +38,22 @@
...
 #' @export detect_pulse
-#' @importFrom data.table setorder
+#' @importFrom data.table setorder,setNumericRounding
 setGeneric("detect_pulse", function(obj){standardGeneric("detect_pulse")})
 
 #' @rdname detect_pulse
 setMethod("detect_pulse", "LAS",
   function(obj)
   {
-    gpstime <- pulseID <- ReturnNumber <- NULL
```
Jean-Romain: redo

R/indentify.r
```diff
@@ -120,14 +111,12 @@ setMethod("detect_flightline", "LAS",
...
 #' @export detect_scanline
 #' @importFrom data.table setorder
-setGeneric("detect_scanline", function(obj){standardGeneric("detect_scanline")})
+setGeneric("detect_scanline", function(obj, dt = 30){standardGeneric("detect_scanline")})
 
 #' @rdname detect_scanline
 setMethod("detect_scanline", "LAS",
   function(obj)
   {
-    gpstime <- scanlineID <- ScanDirectionFlag <- NULL
```
Jean-Romain: redo

floriandeboissieu: I don't remember suppressing the NULL initialisation neither modifying the function detect_scanline (my memory or an old version of the code??), but I should have noticed that when committing, I'll be more careful next time... Thank you for the tip on the NULL initialization, I was wondering why.
For the gpstime uniqueness I had the same doubt, but I checked on all the datasets I had and it seems to have a unique gpstime value for al echos of the same pulse, confirmed by that forum [https://groups.google.com/forum/#!msg/lastools/VWV4DVkiVq8/EIs9Jvjd8KIJ](https://groups.google.com/forum/#!msg/lastools/VWV4DVkiVq8/EIs9Jvjd8KIJ).

Jean-Romain: Make sense, gpstime the moment the laser was shot and not the moment its reflection(s) were registered.

You probably used an old version of the code. Code often changes. Update your fork before changing something is a good practice.

Thanks for this good update.

Jean-Romain:
I corrected an error in your code. Please note to never edit NAMESPACE yourself. It's an automatically written file.

You added an roxygen comment:

```
 #' @importFrom data.table setorder,setNumericRounding
```
No comma is expected. The NAMESPACE was badly generated but you didn't see it because you manually modified the file.

It's definitively not a serious problem. Just keep in mind to do not edit this file yourself. Generate it with `devtools::document()`

PR #33, opened 2016-11-22
Roi query

floriandeboissieu:
roi_query was assigning wrong roi names to output (roinames was assigned to output before aggregation by tiles of extraction which reorders the processing, instead of after). These modifications are fixing this issue assigning names of aggregated groups of queries in the first for loop (replaced by an lapply) and reordering to the original order at the end.


Jean-Romain:
I get it ! It took me a while... Thanks for this commits

PR #34, opened 2016-11-22
Feature: parallelization of roi_query with parLapply.

floriandeboissieu:
I finally chose to separate the case of non-parallel computing. I tested it, all seems to work.

### DIFF
--- a/R/roi_query.r
+++ b/R/roi_query.r
@@ -42,7 +42,8 @@
 #' radius is provided (r2 = NULL) it will extract data falling into a disc.
 #' @param r2 numeric or vector. A radius or a set of radii of plots. If r2
 #' is provided, the selection turns into a rectangular ROI. If r = r2 it is a square obviouly.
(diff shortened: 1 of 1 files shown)


Jean-Romain:
Sounds good thanks you ! I'll merge it tonight. I need to make a couple adjustments before because you forgot some few tricks. It's not important but let me give you two advices for the next time (if there is a next time).

1. Always run `devtool::document()`. It  regenerates the documentation if there are some modifications.
2. Always run `devtool::check()`. It checks the whole package including code, bug, doc etc. Currently it returns 1 warning because of the LASlib C++ code and 2 notes because of the use of GNU Make and the size of the data given as example. More that 1 W and 2 N means you made an error somewhere. Here for example you didn't document the `progress` parameter so at least 1 more note.

Don't worry I'll do it. Thank you for all you commit. I'm glad that somebody tests and uses these functions. I don't use them myself so I didn't test them in a lot of different cases.

ps: I modified the pulseID detection to make it faster.

PR #35, opened 2016-11-23
Fix: roi_index pb with one query, list() added for that case.

floriandeboissieu:
Fix: roi_index pb when only one query is asked, list() added for that case. Works with solo and multiple queries.


Jean-Romain:
I totally trust you because you are the main user of these tools. So I merge without checking the code :smile:

floriandeboissieu:
Thanks, I'm glad to help.

PR #46, opened 2017-02-08
Some changes in lasnormalise

floriandeboissieu:
This is more a draft suggestion than a real pull-request.
As I saw that you considered using a RasterLayer as dtm I think it is useless to pass through lasclassify.
Moreover, in raster::extract there are several ways to interpolate at coordinates (methods, function over buffered points), it would be better if kept available to user in my opinion. I did it in lasnormalise but it could also be done in lasclassify.
A third thing was to have the possibility to keep Zdtm, as it can serve in the following.
Finally, in my opinion NA shouldn't be removed by other than user. I thought of a na.rm argument, but this argument is already being used by raster::extract. Thus I think NA points should be left and filtered by user afterwards.

If you feel some of these suggestions would be good I'll make a real pull-request

### DIFF
--- a/R/lasnormalize.r
+++ b/R/lasnormalize.r
@@ -62,7 +62,7 @@
 #' \link[raster:raster]{raster}
 #' \link[lidR:grid_terrain]{grid_terrain}
 #' @export
-lasnormalize = function(.las, dtm, na.rm=...)
+lasnormalize = function(.las, dtm, keep_Zdtm=FALSE, ...)
 {
   . <- Z <- Zn <- X <- Y <- NULL
 
@@ -80,6 +80,8 @@ lasnormalize = function(.las, dtm, na.rm=...)
     normalized@data$Zground=raster::extract(dtm, normalized@data[,.(X,Y)], ...)
 
     normalized@data[, Z := Z - Zdtm][]
+    if(!keep_Zdtm)
+      normalized@data[, Zdtm := NULL]
 
     gc()
 
--- a/R/lasnormalize.r
+++ b/R/lasnormalize.r
@@ -62,7 +62,7 @@
 #' \link[raster:raster]{raster}
 #' \link[lidR:grid_terrain]{grid_terrain}
 #' @export
-lasnormalize = function(.las, dtm, ...)
+lasnormalize = function(.las, dtm, na.rm=...)
 {
   . <- Z <- Zn <- X <- Y <- NULL
 
@@ -79,12 +79,7 @@ lasnormalize = function(.las, dtm, ...)
 
     normalized@data$Zground=raster::extract(dtm, normalized@data[,.(X,Y)], ...)
 
-    isna = is.na(normalized@data$Zn)
-
-  	if(sum(isna) > 0)
-	    warning(paste0(sum(isna), " points with NA elevation points found and removed."), call. = F)
-
-    normalized@data[, Z := Z - Zground][]
+    normalized@data[, Z := Z - Zdtm][]
 
     gc()
 
--- a/R/lasnormalize.r
+++ b/R/lasnormalize.r
@@ -77,7 +77,7 @@ lasnormalize = function(.las, dtm, keep_Zdtm=FALSE, ...)
   {
     normalized = LAS(data.table::copy(.las@data), .las@header)
 
-    normalized@data$Zground=raster::extract(dtm, normalized@data[,.(X,Y)], ...)
+    normalized@data$Zdtm=raster::extract(dtm, normalized@data[,.(X,Y)], ...)
 
     normalized@data[, Z := Z - Zdtm][]
     if(!keep_Zdtm)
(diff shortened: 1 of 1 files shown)


Jean-Romain:
Hi again,

Many points in this PR

> This is more a draft suggestion than a real pull-request.

So prefer to open a new issue to discuss of that. Moreover you proposed here 5 commits. You should squash them together because it difficult to read them and see the differences with original files.

> As I saw that you considered using a RasterLayer as dtm I think it is useless to pass through lasclassify.

Why ?

> Moreover, in raster::extract there are several ways to interpolate at coordinates (methods, function over buffered points), it would be better if kept available to user in my opinion. I did it in lasnormalise but it could also be done in lasclassify.

I don't understand what do you mean here. The interpolation and the extraction of the values are two different things

> A third thing was to have the possibility to keep Zdtm, as it can serve in the following.

Why not. 

> Finally, in my opinion NA shouldn't be removed by other than user. I thought of a na.rm argument, but this argument is already being used by raster::extract. Thus I think NA points should be left and filtered by user afterwards.

I disagree. A LAS object is always an object which respect ASPRS LAS specifications. Always. Having NA values for a coordinates is not possible. The first reason is a choice of development. LAS object respect ASPRS standard, the second one is technical. lidR is fast also because it skip some tests relative to the integrity of the variables because I know that my functions should always return an object that respect standards

R/lasnormalize.r
```diff
@@ -77,15 +77,14 @@ lasnormalize = function(.las, dtm)
   {
     normalized = LAS(data.table::copy(.las@data), .las@header)
 
-    lasclassify(normalized, dtm, "Zn")
```
Jean-Romain: No. What `lasclassify` does it to call `raster::extract`, but it does it better because it updates the las object by reference

R/lasnormalize.r
```diff
@@ -77,15 +77,14 @@ lasnormalize = function(.las, dtm)
...
+    normalized@data$Zground=raster::extract(dtm, normalized@data[,.(X,Y)], ...)
 
     isna = is.na(normalized@data$Zn)
 
   	if(sum(isna) > 0)
 	    warning(paste0(sum(isna), " points with NA elevation points found and removed."), call. = F)
 
-    normalized@data[, Z := round(Z - Zn, 3)][, Zn := NULL][]
-    normalized = lasfilter(normalized, !isna)
```
Jean-Romain: No, explaination given in my post

R/lasnormalize.r
```diff
@@ -62,7 +62,7 @@
 #' \link[raster:raster]{raster}
 #' \link[lidR:grid_terrain]{grid_terrain}
 #' @export
-lasnormalize = function(.las, dtm, na.rm=...)
```
Jean-Romain: What is that `na.rm = ...` ? I wrote that ? Impossible it is no a valid R code... :confused:

R/lasnormalize.r
```diff
@@ -77,7 +77,7 @@ lasnormalize = function(.las, dtm, keep_Zdtm=FALSE, ...)
   {
     normalized = LAS(data.table::copy(.las@data), .las@header)
 
-    normalized@data$Zdtm=raster::extract(dtm, normalized@data[,.(X,Y)], ...)
+    normalized@data[, Zdtm := raster::extract(dtm, normalized@data[,.(X,Y)], ...)][]
```
Jean-Romain: That's better :-) . But it is exatly what `lasclassify` does. You should avoid duplicated code.

R/lasnormalize.r
```diff
@@ -77,15 +77,14 @@ lasnormalize = function(.las, dtm)
   {
     normalized = LAS(data.table::copy(.las@data), .las@header)
 
-    lasclassify(normalized, dtm, "Zn")
+    normalized@data$Zground=raster::extract(dtm, normalized@data[,.(X,Y)], ...)
```
Jean-Romain: No. What `lasclassify` does it to call `raster::extract`, but it does it better because it updates the las object by reference

floriandeboissieu:
Ok, I have it all wrong this time :-).
I am aware of the weird things that can happen interpolating an interpolation, however I thought it was the way to use a DTM, as DTM is supposed to be a point interpolation and not a surface integration thus not a pixel. Taking the nearest DTM value isn't that a problem when on a slope? It is creating stairs on point height, no?
Anyway, I don't have much experience on this question, so I rely on yours. Cheers

Jean-Romain:
> DTM is supposed to be a point interpolation and not a surface integration thus not a pixel. Taking the nearest DTM value isn't that a problem when on a slope? 

Yes it is. You should have a fine resolution for sloppy terrain. I don't have much experience too. DTM are usually store in GeoTiff files i.e. rasters. Everywhere I looked up the dtm was simply substract like that. I understand your problem with such workflow. The other way is to compute the interpolation for each point. This way you don't need raster data. It's what I call a "true normalization". It easy to implement but more computationally intensive. I plan to add that feature. 

But the interpolation of an interpolation... I don't like that. However if you think I'm wrong I'm ok with that. But you should come with arguments read from somewhere we can trust.

Jean-Romain:
Done, `lasnormalize` allows for normalization without raster

PR #48, opened 2017-02-08
Gap fraction profile

floriandeboissieu:
fix: gap fraction and LAD were returning an error with points under 0
fix: LAD is defined as -log(gap_fraction)/(k*dz) (Bouvier et al. 2015)
add: option z0 to gap_fraction_profile and LAD to choose from where to start the profile

### DIFF
--- a/R/metrics.r
+++ b/R/metrics.r
@@ -139,6 +139,7 @@ stdmetrics = function(x, y, z, i, a, rn, class, pulseID, dz = 1)
 #'
 #' @param z vector of positive z coordinates
 #' @param dz numeric. The thickness of the layers used (height bin)
+#' @param z0 numeric. The bottom limit of the profile
 #' @return A data.frame containing the bin elevations (z) and the gap fraction for each bin (gf)
 #' @examples
 #' z = c(rnorm(1e4, 25, 6), rgamma(1e3, 1, 8)*6, rgamma(5e2, 5,5)*10)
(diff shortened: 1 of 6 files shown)


Jean-Romain:
This is a known bug (actually that was a known bug but I forgot it since a year).

My point was the following one: when reading a file if there are negative values you get a warning: something is wrong in your dataset. At the beginning I filtered negative points in this function. But I change my mind and I thought that the user must provide valid data i.e. positive values (in addition it is specified in the doc).

With a parameter `z0` it solves the problem. Why not. I will check you code (especially L159)

R/metrics.r
```diff
@@ -152,15 +153,14 @@ stdmetrics = function(x, y, z, i, a, rn, class, pulseID, dz = 1)
 #' @references Bouvier, M., Durrieu, S., Fournier, R. a, & Renaud, J. (2015).  Generalizing predictive models of forest inventory attributes using an area-based approach with airborne LiDAR data. Remote Sensing of Environment, 156, 322-334. http://doi.org/10.1016/j.rse.2014.10.004
 #' @seealso \link[lidR:LAD]{LAD}
 #' @export gap_fraction_profile
-gap_fraction_profile = function (z, dz = 1)
+gap_fraction_profile = function (z, dz = 1, z0 = 2)
 {
-  maxz = max(z)
 
-  bk = seq(0, ceiling(maxz), dz)
+  bk = seq(floor((min(z)-z0)/dz)*dz+z0, ceiling((maxz-z0)/dz)*dz+z0, dz)
```
Jean-Romain: you removed `maxz` but you still use it. You means `max(Z)`

R/metrics.r
```diff
@@ -200,15 +201,15 @@ gap_fraction_profile = function (z, dz = 1)
...
+LAD = function(z, dz = 1, k = 0.5, z0 = 2) # (Bouvier et al. 2015)
 {
 	ld = gap_fraction_profile(z, dz)
 
 	if(anyNA(ld))
 	  return(NA_real_)
 
 	lad = ld$gf
-	lad = -log(lad)/k
+	lad = -log(lad)/(k*dz)
```
Jean-Romain: I missed something here?

floriandeboissieu: Yes:
LAI= -log(gf)/k
LAD= -log(gf)/(k*dz)
I made a mistake in first lines of gap_fraction_profile with maxz
I'll fix it, try to devtools::document(), update NEWS.md, devtools::check() and resubmit

R/metrics.r
floriandeboissieu: you saw it!!!

R/metrics.r
Jean-Romain: I checked the paper. Good call

floriandeboissieu: Ok, devtools::document(), update NEWS.md, devtools::check() done
I think I made the new branch from master, that is why it had a conflict with devel in NEW.md, resolved it.

Jean-Romain:
Thank you

PR #82, opened 2017-10-05
Devel pull for lassnags

bi0m3trics:
I think I did this right.... 
Cheers!
Andrew


R/lassnags.r
```diff
@@ -0,0 +1,246 @@
...
+#' page 172. Default is 3}
+#' \item{\code{bbpr_thresholds}}{A 3x4 matrix providing the four average BBPR values for each of the three neighborhood
+#' (sphere, small cylinder and large cylinder neighborhoods) to be used in for conditional assessments and classificaiton
+#' into the following four snag classes: 1) general snag, 2) small snag, 3) vlive crown edge snag, and 4) high canopy 
+#' cover snag. See reference page 172 and Table 2. This matirx must be provided by the user.}
+#' The current implementation is known to use a large amount of memory for storing the N x k integer matrix returning the
+#' near neighbour indices for each point in the point cloud. Improvements are possible in future package versions.}
+#' 
+#' @examples
+#' LASfile <- system.file("extdata", "MixedConifer.laz", package="lidR")
```
Jean-Romain: You can't add you own dataset. The CRAN accepts for 5 MB of data. The limit is already reached. In any cases your code is too slow to be tested by the CRAN. The CRAN tests every example and raise warning for too long examples. You must add a `\dontrun` command in your example. That way the CRAN will skip this example but the user will be able to copy paste it. But not with your dataset.

R/lassnags.r
```diff
@@ -0,0 +1,246 @@
...
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>
+#
+# ===============================================================================
+
+#' Snag classification of airborne LiDAR point clouds using neighborhood metrics
+#'
+#' Snag clasification using the algorithim proposed by Wing et al 2015 (see details). The function
```
Jean-Romain: You created the function with a parameter `algorithm`. So the ultimate goal is not to be restricted to Wing. So far a single algo is implemented. Change a bit how you introduced the function.

I start to rewrite the function `lastrees` because my initial designed was not the best. Try yo get inspiration from `lastree` in the `devel` branch

R/lassnags.r
```diff
@@ -0,0 +1,246 @@
...
+#' attributes to each point of the point cloud a number identifying a snag class (\code{snagCls} column).
```
Jean-Romain: `snagID` would be more constitant with `treeID`, `pulseID`, `flighlineID`. What do you think?

R/lassnags.r
Jean-Romain: Actually you already did that well. Just tweak a bit this first sentence.

R/lassnags.r
```diff
@@ -0,0 +1,246 @@
...
+#' about the algorithm)
+#' 
+#' @return Nothing, the point cloud is updated by reference. A new field is added to the las file (\code{SnagCls}),
+#' for each point with the corresponding classificaiton: 0) live tree, 1) general snag, 2) small snag, 3) live 
+#' crown edge snag, and 4) high canopy cover snag.
+#'
+#' @section Wing et al. 2016:
+#' This is an automated filtering algorithm that utilizes three dimensional neighborhood lidar point-based
+#' intensity and density statistics to remove lidar points associated with live trees and retain lidar points
+#' associated with snags devleoped by Wing et al (2015 - see references).
```
Jean-Romain: devleoped. I can ask Dave to review ;-)

R/lassnags.r
```diff
@@ -0,0 +1,246 @@
...
+#' for each point with the corresponding classificaiton: 0) live tree, 1) general snag, 2) small snag, 3) live 
+#' crown edge snag, and 4) high canopy cover snag.
+#'
+#' @section Wing et al. 2016:
+#' This is an automated filtering algorithm that utilizes three dimensional neighborhood lidar point-based
+#' intensity and density statistics to remove lidar points associated with live trees and retain lidar points
+#' associated with snags devleoped by Wing et al (2015 - see references). 
+#' Note that this algorithm strictly performs a classifciation based on user input while the original publicaiton's 
+#' methods also included a segmentation step and some pre- (filtering for first and single returns only) and post-
+#' process (filtering for only the snag classified points prior to segmentation) tasks whihc are now expected to
```
Jean-Romain: whihc

R/lassnags.r
```diff
@@ -0,0 +1,246 @@
...
+#' crown edge snag, and 4) high canopy cover snag.
+#'
+#' @section Wing et al. 2016:
+#' This is an automated filtering algorithm that utilizes three dimensional neighborhood lidar point-based
+#' intensity and density statistics to remove lidar points associated with live trees and retain lidar points
+#' associated with snags devleoped by Wing et al (2015 - see references). 
+#' Note that this algorithm strictly performs a classifciation based on user input while the original publicaiton's 
+#' methods also included a segmentation step and some pre- (filtering for first and single returns only) and post-
+#' process (filtering for only the snag classified points prior to segmentation) tasks whihc are now expected to 
+#' be preformed by the user. Also, this implemntation may have some differences compared with the original method
```
Jean-Romain: preformed

R/lassnags.r
```diff
@@ -0,0 +1,246 @@
...
+#' due to potential mis-interpretation of the Wing et al. manuscript, specifically Table 2 where they present four
+#' groups of conditional assessments with their required neighborhood point density and average BBPR values (BBPR 
+#' = branch and bole point ratio; PDR = point density requirement). 
+#' 
+#' \describe{
```
Jean-Romain: This the old way I document `lastress`. This was not good. Look at how I do that now in the devel branch.

R/lassnags.r
```diff
@@ -0,0 +1,246 @@
...
+#' \item{\code{neigh_radii}}{A vector of three radii used in quantifying local-area centered neighborhoods. See 
+#' reference page 171 and Figure 4. Defaults are 1.5, 1, and 2 for the sphere, small cylinder and large cylinder neighborhoods,
+#' respectively}
+#' \item{\code{low_int_thrsh}}{The lower intensity threshold filtering value. See reference page 171. Default is 50}
+#' \item{\code{uppr_int_thrsh}}{The upper intensity threshold filtering value. See reference page 171. Default is 170}
+#' \item{\code{pt_den_req}}{Point density requirement based on plot-level point density defined classes. See reference
+#' page 172. Default is 3}
+#' \item{\code{bbpr_thresholds}}{A 3x4 matrix providing the four average BBPR values for each of the three neighborhood
+#' (sphere, small cylinder and large cylinder neighborhoods) to be used in for conditional assessments and classificaiton
+#' into the following four snag classes: 1) general snag, 2) small snag, 3) vlive crown edge snag, and 4) high canopy
```
Jean-Romain: vlive

R/lassnags.r
```diff
@@ -0,0 +1,246 @@
...
+#' (sphere, small cylinder and large cylinder neighborhoods) to be used in for conditional assessments and classificaiton
+#' into the following four snag classes: 1) general snag, 2) small snag, 3) vlive crown edge snag, and 4) high canopy 
+#' cover snag. See reference page 172 and Table 2. This matirx must be provided by the user.}
+#' The current implementation is known to use a large amount of memory for storing the N x k integer matrix returning the
+#' near neighbour indices for each point in the point cloud. Improvements are possible in future package versions.}
+#' 
+#' @examples
+#' LASfile <- system.file("extdata", "MixedConifer.laz", package="lidR")
+#' las = readLAS(LASfile, select = "xyzi")
+#' lasnormalize(las, grid_terrain(las, method = "knnidw", k = 10L))
```
Jean-Romain: Don't normalize. We will suppose the dataset to be already normalized.

R/lassnags.r
```diff
@@ -0,0 +1,246 @@
...
+#' las %<>% lasfilter(Z>1.37, ReturnNumber==1)
```
Jean-Romain: Use streaming filter in readLAS. It's fine since we don't normalize anymore

R/lassnags.r
```diff
@@ -0,0 +1,246 @@
...
+#' lassnags(las, algorithm = "wing2015", neigh_radii = c(1.5,1,2), low_int_thrsh = 50, uppr_int_thrsh = 170, pt_den_req = 3, bbpr_thresholds = BBPRthrsh_mat)
+#'     
+#' # Plot it all, tree and snag points...
+#' plot(las, color="SnagCls", colorPalette = rainbow(5))
+#'     
+#' # Filter and plot snag points only 
+#' las %<>% lasfilter(SnagCls>0)
+#' plot(las, color="SnagCls", colorPalette = rainbow(5)[-1])
+#' 
+#' Preform tree segmentation using the watershed method
```
Jean-Romain: Why? Explain or remove

R/lassnags.r
```diff
@@ -0,0 +1,246 @@
...
+#' chm = grid_canopy(lasnorm, res = 0.5, subcircle = 0.2, na.fill = "knnidw", k = 4)
+#' chm = as.raster(chm)
+#' lastrees(lasnorm, "watershed", chm, th = 4)
+#'  
+#' @references
+#' Wing, Brian M.; Ritchie, Martin W.; Boston, Kevin; Cohen, Warren B.; Olsen, Michael J. 2015. 
+#' Individual snag detection using neighborhood attribute filtered airborne lidar data. Remote 
+#' Sensing of Environment. 163: 165-179 https://doi.org/10.1016/j.rse.2015.03.013\cr\cr
+#'
```
Jean-Romain: Add @author with your name and mine. For all the other function it was implicitly me. But here it is not.

R/lassnags.r
```diff
@@ -0,0 +1,246 @@
...
+  {
+    idx = nn_idx[i, ] 
+    lg_cyl_BranchBolePtRatio_mean[i] = mean(lg_cyl_BranchBolePtRatio[idx])
+  }
+  
+  # ====== STEP 2 =======
+  # Point classificaitons based on rough interpretation of Table 2 - pg. 172
+  # Vlaues supplied/specified by user in BranchBolePtRatio.mat
+
+  las@data[, SnagCls := ifelse(sph_PtDen>=pt_den_req  &
```
Jean-Romain: Could you rewrite that with intermediate steps to make it more understandable

tests/testthat/test-lassnags.R
```diff
@@ -0,0 +1,17 @@
+context("lassnags")
```
Jean-Romain: Idem that's a too long computation. CRAN will complain.

R/lassnags.r
bi0m3trics: Ok. I understand, however all of the existing example datasets either don't have snags (e.g., Megaplot.laz) or have meaningless intensity values (e.g., Tree.laz). I can add \dontrun but this still means there's no meaningful example. I could decimate a small dataset with a snag or two and provide that... thoughts? Otherwise reading a dataset without snags or intensity will just be confusing...

Jean-Romain: I agree, this is a problem. Is your dataset suitable for a good tree segmentation? We could replace the dataset used for tree segmentation example by your dataset. But you must find a way to have a file of the same size.

bi0m3trics: I'll play with a few datasets and find something that suits all scenarios, is the same size, and is easy to document. I'll then send it to you separately and let you decide.

R/lassnags.r
bi0m3trics: Holy cow - that's awesome and I completely missed it!  ;)

Jean-Romain: You're not the only one. This is why I try to emphasis it in the doc

R/lassnags.r
bi0m3trics: At present, it's just a classification algorithm (wing et al), so I left it as snagCls but included in the doc the possibility of attributing snagID in future algorithms

tests/testthat/test-lassnags.R
bi0m3trics: Should I just delete it then?

Jean-Romain: The dataset you send me by email is nice. Compressed in laz its weight dropped to 500 kB and the tree segmentation is easy and therefore the data is suitable in replacement of the current file used as example `Tree.laz`.

However it is still too big for two reasons:
* The total weight of example files is greater than 5 MB
* The time spent of the `lastrees` example (using li 2012) is too long and the time spent on `lassnag` is too long too.

The solution is to decimate your dataset to 5 pt/m² and crop it a bit. Thus the dataset is still good for an example of tree segmentation (good sementation in less than 1 second) and snags detection (output is almost the same after decimation). 

But `lassnags` is still too slow to be tested by the CRAN. So the example must be wrapped in `\dontrun`. And the test in `testthat` must be removed (commented only, thus if we speed up the algo we can simply uncomment).

And you must review your example. It does not work. Also run `devtools::check()` before a PR to check everything. You should have 0 error, 0 warning, 0 notes (actually the devel branch have currently some warnings. Check only those which concern `lassnags`)

R/lassnags.r
```diff
@@ -218,28 +219,29 @@ lassnags_wing = function (las, neigh_radii = c(1.5,1,2), low_int_thrsh = 50, upp
...
-  # Vlaues supplied/specified by user in BranchBolePtRatio.mat
+  # values supplied/specified by user in bbpr_thresholds
 
   verbose("Classifiying points...")
+  las@data[, snagCls :=
   ifelse(sph_PtDen>=pt_den_req  & sph_BranchBolePtRatio_mean>=bbpr_thresholds[1,1] &
          sm_cyl_PtDen>=pt_den_req & sm_cyl_BranchBolePtRatio_mean>=bbpr_thresholds[2,1] &
          lg_cyl_PtDen>=pt_den_req & lg_cyl_BranchBolePtRatio_mean>=bbpr_thresholds[3,1],
-         las@data[, SnagCls := 1],        # General snag class
+         1,        # General snag class
```
Jean-Romain: I can't see any reference in the doc to the labels used. Please document 1 = general snag and so on

tests/testthat/test-lassnags.R
bi0m3trics: Gotcha... I'm on it asap

R/lassnags.r
bi0m3trics: I'll get this in the documentation. Should be easy enough...

Jean-Romain:
Sounds good to me. I will make final modifications myself.

Last question but not the least. How confident you are on the fact that you have really implemented Wing's algo? How did you ensure the results are good?

I did not read Wing et al. so you can implement anything I must trust you. However I corrected some mistakes in your code without reading the original paper. I can't see obvious mistake anymore but I would like to be sure this is really the Wing's algo. I perfectly know that some part of the algo may be difficult to understand and sometime when implementing an algo we have to make some guesses and interpretations. But the question remains. How could you convince me this is a fair implementation of Wins's algo?

bi0m3trics:
Before I answer that last question....  I still cant get my example to run....not because it has errors but because readLAS keeps returning this error:
`Error: Incorrect argument 'filter'`
I know what I've written is right, but I think it's something to do with my PR being in devel so I couldn't check the example for time with the streaming filer added... I did however check it with an older verison and it ran fine.

Jean-Romain:
Filter is fine on my computer. However the result itself is no more the same than this afternoon. These is no snags anymore. The point cloud is red.

bi0m3trics:
I completely understand your question regarding implementing Wing's algorithm... 

I stuck to it as closely as possible and I think I've made it clear that the Table 2 part (the classifications) is a little wonky as written (you have to read it and try to implement it to catch it). In the beginning, I worked hard with my old, data.table heavy code to make sure it was all working correctly (hand calculating the final  neighborhood BBPR values for numerous points as checks) and somewhere around version 2 or 3 (about the time you told me that `mean` was "useless" and returned a scalar is when I realized something had gotten off. I went back though and by the time you said you weren't editing anything anymore I had recalculate values by hand and confirmed that they were right. Simply put, its the best and closet to what I read his paper, and I take full responsibility if anything was mis-interpreted. 

Lastly, the only thing that I allowed to remain different is that he filters for both first and single returns, and if I did this the snags would have only been 1 point each (i.e., there's just not enough points in the example)

bi0m3trics:
> The point cloud is red. 
- even after running:
```
las %<>% lasfilter(snagCls>0)
plot(las, color="snagCls", colorPalette = rainbow(5)[-1])
```
There should be no red as the first color is red and being removed? Once I run this...I only see 4-5 snags and they're all yellow and blue. Let me screen grab it...

Jean-Romain:
Remains 2 points only. This afternoon I got 2 or 3 yellow-blue trees.

bi0m3trics:
ah... I got it now... this is because of the last line of the "did I implement it correctly part..."
If you apply all three streaming filters (-drop_z_below 0 -keep_first -keep_single) you only get a few points... as there's nothing left to show (just a few points). 
If you only apply `-drop_z_below 0 -keep_first` it will look like this:
filtersnagpointsonly
This last fix will be quick...so my next push to origin should look like that capture.
2 seconds

bi0m3trics:
Ok. Now when you run the example it should look just like that capture and my last line in the earlier comment stands.
> Lastly, the only thing that I allowed to remain different is that he filters for both first and single returns, and if I did this the snags would have only been 1 point each (i.e., there's just not enough points in the example)

bi0m3trics:
I also see why you asked now  ;)

bi0m3trics:
Once you run it and let me know if you got the same thing... let me know if you think I should edit the example further to let the user know what filters are applied by Wing and what's being applied in this example due to file size limitations.

Jean-Romain:
I got the same display as you and the same as this afternoon.

For the documentation you're are free to write whatever you want. I'm only checking if the doc meets CRAN requirements (still parameter `extra` useless and undocumented) and if everything is documented. But I do no check if the explanations are exhaustive regarding Wing's paper.

> and I take full responsibility if anything was mis-interpreted.

Unfortunately no, you don't. I mean, for sure you take the responsibility if something is wrong and I do not doubt of you. But in practice, users will just use wrong results without assuming it could be wrong. Users usually just trust in the software.

bi0m3trics:
Well that's good that you got the same thing. Hopefully the filtering thing makes sense after we decimated the final example data set. That was just an oversight on my end...

Regarding responsibilities, sometimes it's unfortunate that you can't fix ethics problems with software! I say trust nothing! ;)

Thanks for all your help again and I hope you have a good weekend. Do graduate students get weekends?

Sent from my Verizon 4G LTE smartphone


-------- Original message --------

Jean-Romain:
> Do graduate students get weekends?

Not those who do software development in addition to their PhD :wink: .

Jean-Romain:
What about L131? First is was an `or` then I replaced by `length` then you changed for a `and` and now it is a `length`.

Jean-Romain:
```r
sum(sph_BranchBolePtRatio[idx])/length(idx)
```

Instead of mean. I gained 20% again

Jean-Romain:
```r
Int_vec[idx]
```

since `idx` often contains several 0 what are the consequences on the computation?

bi0m3trics:
L131 is right. I think I changed and then changed it bak too... In the paper Wing had just written it two ways (theoretical and how it was implemented). The way I orginally implemented in the beginning resulted in ratios > 1 and upon re-reading it he did have the denominator as the points  in each neighborhood. Lucky maybe ;)



Sent from my Verizon 4G LTE smartphone


-------- Original message --------

bi0m3trics:
Never would have thought mean had that much overhead. Must be the check for NAs, which we don't need.



Sent from my Verizon 4G LTE smartphone


-------- Original message --------

bi0m3trics:
Ok. This was the tough one that I know I "settled" on... the zeros are not needed but nn2 didn't return variable length rows and removing the zeros after seemed computationally expensive since returning index zero would be fast... maybe not. In short, each neighborhood has a different number of idx up to an estimated k, but the non-zeros in each row are all that are actually needed. I'd welcome and suggestions and be happy to check any solution.



Sent from my Verizon 4G LTE smartphone


-------- Original message --------

Jean-Romain:
> Never would have thought mean had that much overhead. Must be the check for NAs, which we don't need.

Yes if  `na.rm = TRUE` otherwise i'm not sure.. But `mean` does much more than mean.

>  removing the zeros after seemed computationally expensive [...] but the non-zeros in each row are all that are actually needed. I'd welcome and suggestions and be happy to check any solution.

I will make some tests. I guess I can make optimizations but I need to ensure I will not break anything. This why I asked you "what really happens with the 0". I'm surprised there not even a warning.

PR #89, opened 2017-11-16
Extra bytes

floriandeboissieu:
Corresponding changes to rlas EXTRA BYTE support.
I let you check, and if it is OK wth you add a line in the news.md.
Cheers,
Florian


Jean-Romain:
The new version of rlas must be compatible with lidR without this patch. We must find the best compatible implementation in rlas first without changing lidR. Then we will enable the new features in lidR.

Jean-Romain:
What do think about using `boost::any`. Sounds to be the best solution.

PR #92, opened 2017-11-17
Changes to satisfy clang++ compiler

vlulla:
Fixes the following various ambiguous type errors generated by clang++ (3.9.1) on mac:

```
algo_li2012.cpp:57:13: error: conversion from                                                  
      'Rcpp::SlotProxyPolicy<Rcpp::S4_Impl<PreserveStorage> >::SlotProxy' to 'DataFrame' (aka  
      'DataFrame_Impl<PreserveStorage>') is ambiguous                                          
  DataFrame data = las.slot("data");                                                           
            ^      ~~~~~~~~~~~~~~~~                                                            
/Library/Frameworks/R.framework/Versions/3.4/Resources/library/Rcpp/include/Rcpp/proxy/SlotProx
y.h:42:31: note:                                                                               
      candidate function [with T = Rcpp::DataFrame_Impl<PreserveStorage>]                      
        template <typename T> operator T() const;                                              
                              ^                                                                
/Library/Frameworks/R.framework/Versions/3.4/Resources/library/Rcpp/include/Rcpp/DataFrame.h:51
:9: note:                                                                                      
      candidate constructor [with T = Rcpp::SlotProxyPolicy<Rcpp::S4_Impl<PreserveStorage>     
      >::SlotProxy]                                                                            
        DataFrame_Impl( const T& obj ) ;  
```

Slight changes to remove `-Wnewline-eof warnings! `

### DIFF
--- a/src/BoundingBox.h
+++ b/src/BoundingBox.h
@@ -14,3 +14,4 @@ struct BoundingBox
 };
 
 #endif //BBOX_H
+
--- a/src/Point.h
+++ b/src/Point.h
@@ -41,3 +41,4 @@ private:
 };
 
 #endif //POINT_H
+
--- a/src/QuadTree.h
+++ b/src/QuadTree.h
@@ -44,3 +44,4 @@ class QuadTree
 };
 
 #endif //QT_H
+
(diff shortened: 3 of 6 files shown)


Jean-Romain:
Hi thank you for this PR.

You're right this was ambiguous but gcc does not complained so I kept it as is so far. 

I have a question. Are you sure is does not create a deep copy? Actually I don't know the behavior of `as` in `Rcpp` when used from type A to type A. I usually used it to transtype R object to STL object.

Jean-Romain:
I tested it myself and it is ok. Thank you for the correction

PR #102, opened 2018-01-24
Write extra byte

floriandeboissieu:
writeLAS interface for writting extra bytes.
It works on my platform, but not on rhub check maybe linked to rlas compilation pb with clang on rhub check, see rlas PR.


Jean-Romain:
I close this PR. Everything works in `lidR 1.5.0` and is exepected to be driven and controlled by `rlas 1.2.0`. See also #113

PR #104, opened 2018-02-05
Update function 'entropy' in 'metrics.R'

frank2165:
breaks calculation in the function 'entropy' has the potential to exclude points when supplied bin-width argument is greater than 1. For example when zmax = 20.38 and by = 5 the last break point will be 20 and the maximum point will be excluded, returning an error. The proposed change uses the same method as implemented for a similar calculation in the function 'gap_fraction_profile'.


Jean-Romain:
You're right. Thanks for catching that.

PR #156, opened 2018-07-19
Update function lasclip.LAS in lasclip.r

frank2165:
Variable `las` is not created when argument `geometry` is a `SpatialPolygonsDataFrame` causing error.


Jean-Romain:
Nice catch. Thanks.

PR #159, opened 2018-07-23
Fix bug introduced into cluster_apply

frank2165:
Tests of the latest version of catalog_apply/cluster_apply on a more complicated function did not adequately capture all of the required packages. The code I wrote additionally fails when the name attribute of each element of `where` returns `NULL`, as `vapply` expects a length one character vector. 

It may also be worthwhile adding the condition that these checks are only run when the machine does not support multicore processing, i.e. changing the condition to `ncores > 1 && !future::supportsMulticore()`, so that non-windows users aren't unnecessarily impacted by these checks. I haven't added this code into the pull request as I can't test what this inclusion would do outside of a windows environment (I've broken enough code for one day).

### DIFF
--- a/R/clusters_apply.r
+++ b/R/clusters_apply.r
@@ -10,7 +10,7 @@ cluster_apply = function(clusters, f, ncores, progress, stop_early, ...)
   future::plan(future::multiprocess, workers = ncores)
 
   # User supplied function not being analysed for globals/packages by the future we have to do it manually.
-  if (ncores > 1)
+  if (ncores > 1 && !future::supportsMulticore())
   {
     dots <- list(...)
     is.fun <- vapply(dots, is.function, logical(1))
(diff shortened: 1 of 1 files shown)


Jean-Romain:
> `!future::supportsMulticore()` so that non-windows users aren't unnecessarily impacted by these checks.

Great I was look for something like that but i did not find the good functions.

> The code I wrote additionally fails when the name attribute of each element of where returns NULL, as vapply expects a length one character vector.

Could you show me a simple reproducible example. We have to check really carefully all that stuff. The `lidR` package has many unit test but all run on a single core so that I have to test manually multicore options.

frank2165:
(Un)fortunately I haven't been able to reproduce the case where `attr(where, "name")` returns NULL and breaks the vapply statement, I must have just done something dumb.

As for the remaining changes, the following function fails for the current package version (`Error in summarise(., MaxZ = max(Z), MaxCount = max(Count), HtMaxCount = Z[which.max(Count)]): could not find function "summarise"`):

```r
library(lidR)
library(dplyr)
library(magrittr)

las <- catalog(dataDIR)
cores(las) <- 3

test_fun <- function(las){
  grid_metrics3d(las, list(Count = length(Intensity))) %>%
    summarise(MaxZ = max(Z), MaxCount = max(Count), HtMaxCount = Z[which.max(Count)])
}

catalog_apply(las, test_fun)
```

This is occurs because `future::getGlobalsAndPackages` finds where the `summarise` function is defined (`package:dplyr`) and places the package name in the "packages" element of the list that it returns. I was hasty in writing the original code and only handled the case where functions show up in the "globals" element of the output, like magrittr's pipe operator does. This is the rationale behind adding the following code `required.pkgs <- c(required.pkgs, setdiff(globals$packages, required.pkgs))` to ensure that all of the required packages are captured.

An additional problem that I hadn't considered initially is that `future::getGlobalsAndPackages` will find globals in the environment `imports:lidR` and report these in the output (see code below). If the `imports:<package-name>` environments are not filtered out then the function will produce the following error: `Error in loadNamespace(pkg) : there is no package called ‘imports:lidR’`. This is fixed by adding `pkgs    <- unique(grep("package\\:", pkgs, value = TRUE))` to ensure that only valid package names are provided in the call to future.

```r
test_fun_2 <- function(las){
  as.spatial(las) %>% as("SpatialPoints")
}
catalog_apply(las, test_fun_2)
```

PR #161, opened 2018-08-03
bug in treetops ID check

floriandeboissieu:
There was a bug on treetop ID.
However, it raises a question about the expected third column, which seems to be an ID and not a Z value as it is given by tree_detection. Thus if several tree tops have the exact same height, the algorithm throws an error. And if several tops have the same integer part they will be identified with the same color. Thus
Thus, to make it work properly without the user-defined ID option (like in the other algorithms):
- I replaced treetops_df[, 3] by  treetops_df[[3]]
- I changed to generate a unique treetops_df integer ID systematically
If the possibility to have a user-defined treetop ID is wanted, it could be either given in a new argument or expressed specifically to be the third column or the RasterLayer value in the manual. In that case, this option should be thought for the other algorithms as well, to be homogeneous and to keep lastrees working, shouldn't it?

### DIFF
--- a/R/lastrees_dalponte.r
+++ b/R/lastrees_dalponte.r
@@ -17,7 +17,7 @@
 #' @param chm RasterLayer. Image of the canopy. Can be computed with \link[lidR:grid_canopy]{grid_canopy}
 #' or \link[lidR:grid_tincanopy]{grid_tincanopy} or read it from an external file.
 #' @param treetops \code{RasterLayer} or \code{data.frame} containing the position of the
-#' trees. Can be computed with \link[lidR:tree_detection]{tree_detection} or read from an external file.
+#' tree tops, with X and Y as first and the second column respectively. Can be computed with \link[lidR:tree_detection]{tree_detection} or read from an external file.
 #' @param th_tree numeric. Threshold below which a pixel cannot be a tree. Default 2.
 #' @param th_seed numeric. Growing threshold 1. See reference in Dalponte et al. 2016. A pixel
 #' is added to a region if its height is greater than the tree height multiplied by this value.
@@ -66,20 +66,15 @@ lastrees_dalponte = function(las, chm, treetops, th_tree = 2, th_seed = 0.45, th
   assertive::assert_all_are_in_closed_range(th_cr, 0, 1)
 
   if (is(treetops, "data.frame"))
-  {
-    treetops_df = treetops
-
-    if (ncol(treetops_df) < 3)
-      treetops_df$id = 1:nrow(treetops_df)
-  }
+    treetops_df = treetops[,1:2] # keep only XY coordinates
   else if(is(treetops, "RasterLayer"))
-    treetops_df = raster::as.data.frame(treetops, xy = TRUE, na.rm = TRUE)
+    treetops_df = raster::as.data.frame(treetops, xy = TRUE, na.rm = TRUE)[,1:2]
   else
     stop("'treetops' format not recognized.", call. = FALSE)
-
-  if (length(unique(treetops_df[, 3])) != nrow(treetops_df))
-    stop("Duplicated seed IDs.", call. = FALSE)
-
+  
+  # set tree top ID
+  treetops_df[[3]] = 1:nrow(treetops_df)
+  
   cells = raster::cellFromXY(chm, treetops_df[,1:2])
 
   if (anyNA(cells))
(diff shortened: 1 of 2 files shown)


biometrica:
JR,

Has anyone ever comments about the amazing similarity in outputs from
Dalponte and Silva?

I ran these algorithms over 350-stands and the outputs for implied stocking
(trees/acre) were almost identical.

*Nate Osborne, Ph.D.*
Biometrics Project Leader
1 Rayonier Way, Wildlight, Florida, 32097
( 904-763-4934

On Fri, Aug 3, 2018 at 11:25 AM, floriandeboissieu <notifications@github.com

floriandeboissieu:
See second and third commits on outside CHM seed fix and manual update

Jean-Romain:
@floriandeboissieu  did you run all unit tests ? I suspect, reading very quickly your code, that the ID of the trees is no longer preserved. Right ?

Also I'm currently working on the new version [2.0 of lidR](https://github.com/Jean-Romain/lidR-v2.0). The bug you corrected will not be relevant anymore. lidR 2.0 will be way different !
But it does not matter id there is a bug in version 1.6.0 I will release a version 1.6.1 with some fixes.

@biometrica They are not strictly equivalent but:
- They are both not really advanced growing region algorithm and I suspect you could have almost the same result just removing low pixels
- You used the same seeds for both methods. There are both growing region methods.  

Consequently they **do provide** exactly the same trees, with very minor variation in the delineation. I'm not really surprised.

biometrica:
@Jean-Romain Thanks for the direction on that.

Jean-Romain:
As I guessed it does not pass tests:

```
test-lastrees.R:33: warning: Dalponte's methods works
Seeds outside the canopy height model are removed.

test-lastrees.R:33: failure: Dalponte's methods works
`lastrees_dalponte(las, chm, ttopsdf, extra = T)` did not throw an error.

test-lastrees.R:40: failure: Dalponte's methods works
sort(unique(seg3[])) not equal to 1:59 * 2.
59/59 mismatches (average diff: 30)
[1] 1 -  2 == -1
[2] 2 -  4 == -2
[3] 3 -  6 == -3
[4] 4 -  8 == -4
[5] 5 - 10 == -5
[6] 6 - 12 == -6
[7] 7 - 14 == -7
[8] 8 - 16 == -8
[9] 9 - 18 == -9
...

I do agree that error may be transformed to warning (test must be changed). However ID must be preserved.

**Comment by floriandeboissieu** [comment]
I did not run the test, sorry, I did not have the time.
I found the bug running something like:
```
LASfile <- system.file("extdata", "MixedConifer.laz", package = "lidR")
las = readLAS(LASfile, select = "xyzr", filter = "-drop_z_below 0")
ttops = tree_detection(las, 3, 2)
lastrees_dalponte(las, chm, ttops)
```
which gave an error because treetops_df[,3] is a column of a data.table and not a vector.
I fix it with treetops_df[[3]], but the result was not coherent.

It seems that there are two different behaviors giving two different results depending on if you use chm or las at tree_detection step:
- with chm as input tree_detection gives a RasterLayer with IDs of tree tops,
- with las as input tree_detection gives a data.table with XYZ columns
I used the second frame, the 3rd column being a Z value that is then round to give an ID in the C_dalponte code. Thus similar treetops height becomes same ID.
That is why I removed the ID column that can be easily retrieved a posteriori in the case of the data.frame input. But if you wish to keep the ID the simplest solution is to specify that a third column is expected as ID in the manual, and give an example starting with tree_detection on las that would change last column of ttops to IDs. In that case, only accept the warning and the treetops[,3] to treetops[[3]].

However, I don't really understand why you would want a raster full of NA as output tree_detection (I don't have any example in mind for which it would be usefull). I would better give a unique SpatialPointsDataFrame output independently of input (chm or las) to keep projection in both cases. In that case, at segmentation step, ID could be either given in a separate argument (e.g. idcol like in data.table::rbindlist) or left to the user to compute it on its own.

By the way, very nice initiative the v2.0, I was thinking exactly the same thing about data.table outputs.

Jean-Romain:
> I would better give a unique SpatialPointsDataFrame

This what I did for v2.0

> However, I don't really understand why you would want a raster full of NA as output tree_detection
Short awnser [(long answer here)](https://github.com/Jean-Romain/lidR-v2.0)

The lidR package was initially built upon "personal R scripts" that I wrote during my PhD. These scripts were written for my own use [...] at the time I had no plans to build a big framework. [...]

The package is now a relatively big framework [...] built upon a perfectly unstructured base. It started simply as a collection of R scripts for personal use, and any additional functionality was appended in a very unorganized manner.

So in short, lidR is like a large building that is built on top of a very tiny house! I cannot improve the package further without building a new strong base and that implies breaking many things. And this is exactly what I will do. I will break things in v2.0 to make something coherent and structured !

I will fix this bug tomorrow. Also you can

- Discuss about all the incoherence of the current package design [here](https://github.com/Jean-Romain/lidR-v2.0) 
- Follow the work-in-progress [here](https://github.com/Jean-Romain/lidR/blob/v2/NEWS.md)

Jean-Romain:
Fixed in a dirty way but I'm now working on v2.0 that is **muuch** more robust and coherent will all other geospatial tools in R.

PR #189, opened 2018-11-11
Devel - Lassnags PR

bi0m3trics:
Once again, I don't do these pull requests enough so if I did something wrong just let me know. This request updates the lassnags code to use Rcpp, lidR's QuadTree search functionality, and eliminates the dependency on RANN.


Jean-Romain:
I quickly read your code. I can't see any obvious problem or improvement. It looks really great!

I will inspect that deeper within few days and I will also test if there are possible memory optimizations but I don't think so. Thank you for your work this will greatly improve your original algorithm. To make a full PR you can also remove the dependency to RANN in the file DESCRIPTION

If only have one question. Why the algo does no return strictly the same output as the former version. You changed a test into `expect_equal(as.numeric(table(las@data$snagCls)), c(7223, 93, 55, 15))`. Do you think you replaced a `<` by a `<=` or the opposite somewhere?

bi0m3trics:
I racked my brain for the better part of a couple days trying to figure out why the numbers didn't match in the test, and all I can come up with is either 1) the old code was wrong and I used the wrong n when calculating the denominator for the small cylinder mean BBPr (there's no if Z >= the focal point Z check in the old code since I used nn2 indices), or 2) maybe it just had to do with my use of the RANN::nn2 approach to get around the small and large cylinder neighborhood object calculations... I know the gigantic classification statement is right as that code translated (copied, really) easily. Ultimately, I'm going with this code as correct. Let me know if you see anything when you're tweaking things... I'll commit the DESCRIPTION change here in a bit.

Jean-Romain:
Merged. Everything sounds good to me I did not find any error. The runtime on the example file has been divided by ~3.4  and the memory usage by ~75. Great! Now you're a C++ pro :smiley: .

I can point out few things anyway:

* You should really use you space bar when you code :wink: 
* In the former version you wrote

```r
pcPtDen = las@header@PHB$`Number of point records`/area(las) # The point cloud point density (per m)
[...]
k = ceiling(pcPtDen*pi*r1^2/10)*10 # The maximum number of neighbors to be included
nn_idx = RANN::nn2(XYZ, XYZ, k = k, treetype = "kd", searchtype = "radius", radius = r1)$nn.idx
```

And that does not appear in the new code. You considered the entire shape neighborhood without limits on `k`. I think this is the reason you have different outputs.

* I made few modifications in your code (I added space everywhere mainly)

bi0m3trics:
Agreed on the neighborhood observation and I'll try and use the space bar a little more next time! 😉 but if that's your only criticism, I'll mark this one down as a huge success.


Thanks again for your help on this one. It was a great learning (or reminder) experience and I have a new bat roost (snag) project that's been waiting for me to get this one done... I looked and next year will be 20yrs since I took CS I & II! Maybe I'll take a stab at the "approximately coplaner" classification for identifying man-made structures this winter... that and some wildlife habitat-voxel related work I've been toying with!


On an unrelated note, I've been working with 2.0.0 quite a bit and I must say I've encountered little problems. I've found quite a few typos (more likely English translations) and I'll begin compiling those into a separate PR.


Thanks again!

-bi0m3trics

Jean-Romain:
> if that's your only criticism, I'll mark this one down as a huge success.

The only one. 

> Agreed on the neighborhood observation

Do you need a limited knn function like that `tree->knn_limited(pts, k, shape);`

>  I've found quite a few typos (more likely English translations) and I'll begin compiling those into a separate PR

Dave is currently reviewing everything. Don't worry with that

bi0m3trics:
Regarding the limited knn function -  your knn function is meeting my needs at present, but I could see wanting to limit the search in a sphere or a cube for future applications.  So yes... but only if it's not getting in the way of other enhancements.

PR #257, opened 2019-06-07
Minor corrections to written examples

cjber:
I have modified the code to ensure a working example.


Jean-Romain:
Thanks for the fix.

PR #312, opened 2020-01-12
Fedora packages added

neteler:
As lidR works nicely on Fedora, respective `dnf install ...` line added for convenience.


Jean-Romain:
Thank you

PR #334, opened 2020-03-25
use sf::st_geometry() to get geometry column from sf object

mikoontz:
using the st_geometry() function instead of $geometry ensures that the clipping functions work when the geometry column is named something other than "geometry"

See Issue 333: [https://github.com/Jean-Romain/lidR/issues/333](https://github.com/Jean-Romain/lidR/issues/333)

### DIFF
--- a/R/clip_roi.R
+++ b/R/clip_roi.R
@@ -385,7 +385,7 @@ clip_sf = function(las, sf)
 
 clip_sf.LAS = function(las, sf)
 {
-  wkt <- sf::st_as_text(sf$geometry, digits = 10)
+  wkt <- sf::st_as_text(sf::st_geometry(sf), digits = 10)
 
   output = vector(mode = "list", length(wkt))
   for (i in 1:length(wkt))
@@ -405,7 +405,7 @@ clip_sf.LAS = function(las, sf)
 
 clip_sf.LAScatalog = function(las, sf)
 {
-  wkt  <- sf::st_as_text(sf$geometry, digits = 10)
+  wkt  <- sf::st_as_text(sf::st_geometry(sf), digits = 10)
 
   bboxes <- lapply(wkt, function(string)
   {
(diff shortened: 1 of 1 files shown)


Jean-Romain:
Interesting I didn't know that. I'm not a `sf` user. There is at least one other occurrence here: https://github.com/Jean-Romain/lidR/blob/f5abd06d5a7b90e41044e61ebe4d51f9b08e6253/R/lasmergespatial.r#L198

I can't merge this PR because you made the change from `devel`. This is a bug fix that should be put in `master` first. Sadly I can't merge easily in `master` because I made a lot of change in `v3.0.0` including renaming many files. It gonna be a mess. Please make another PR from `master` to `master`. Then, I'll be able to merge `master`  in `devel`. But not the opposite.

Thank you for the fix anyway :wink:

Jean-Romain:
In `master` the file is `lasclip.r` btw

mikoontz:
Funny, I found this first while using the devel branch (v3.0.0), but did notice after submitting this PR that it is also a bug in the master branch (v2.2.3).

I'm aware of the overhaul of the package (including lots of name changes) between v2.2.3 and v3.0.0. I'm hoping to use the devel  branch with this patch applied.

Would it be helpful for me to submit an independent PR in the master branch with a similar patch in the lasclip.r file? Would you be able to merge each PR separately into devel and into master so the patch can be used in both v2.2.3 and v3.0.0?

mikoontz:
I went ahead and submitted a PR for the master branch (v2.2.3): [https://github.com/Jean-Romain/lidR/pull/335](https://github.com/Jean-Romain/lidR/pull/335)

If it's possible to merge that PR into master, and this PR into devel, that would be awesome!

Thanks so much for your quick response on this.

Jean-Romain:
No, merging more complex that. It will create conflicts :wink:

mikoontz:
Okay! You must be a git magician to merge that into master and then to merge into the dramatically different devel codebase! I defer to your expertise. Thanks for merging the other PR.

PR #335, opened 2020-03-26
fixes bug where simple features with geometry column not named...

mikoontz:
..."geometry" would throw an error

Using sf::st_geometry() to access the geometry column of a simple feature ensures the geometry column (whatever it may be called) is used to get the geometry. If $geometry is used and the geometry column is called something different (e.g., lots of real-world instances where geometry column is called "geom") then NULL results, and sf::st_as_text() throws an error.

### DIFF
--- a/R/lasclip.r
+++ b/R/lasclip.r
@@ -332,7 +332,7 @@ lasclipSimpleFeature = function(las, sf)
 
 lasclipSimpleFeature.LAS = function(las, sf)
 {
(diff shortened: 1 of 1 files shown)


Jean-Romain:
Thank you for the fix. Merged in `master` and then in `devel`. Everything is up-to-date.

mikoontz:
Right on! Thank you so much!

PR #345, opened 2020-05-15
typo that prevents install

mikoontz:
\item instead of \iten, which causes an error upon install with {remotes}

```
remotes::install_github("Jean-Romain/lidR@devel")

...
Error : (converted from warning) /private/var/folders/xl/hy7rwmvx4llff5_00gwbk2vh0000gp/T/RtmpfKPFkY/R.INSTALLe5de7a1086de/lidR/man/LAScatalog-class.Rd:116: unknown macro '\iten'
ERROR: installing Rd objects failed for package ‘lidR’

...
```

### DIFF
--- a/man/LAScatalog-class.Rd
+++ b/man/LAScatalog-class.Rd
@@ -113,7 +113,7 @@ allowed templates is described in the documentation for each function. See \link
 \item \strong{drivers}: list. This contains all the drivers required to seamlessly write Raster*,
 Spatial*, LAS objects. It is recommended that only advanced users change this option. A dedicated
 page describes the drivers in \link{lidR-LAScatalog-drivers}.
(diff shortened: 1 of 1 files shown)


mikoontz:
I did just confirm that `remotes::install_github("mikoontz/lidR@patch-1")` successfully installs.

Jean-Romain:
Thank you for reporting. Sadly I cannot merged your PR I made the change myself. You only changed the Rd file. Rd files are automatically generated. The next documentation rendering will erase your change. The good way is to change the roxygen documentation in R files :wink: 
Best.

mikoontz:
doh! That makes more sense; thanks so much. Next time I'll change the .R and then let roxygen change the rest upon build. I appreciate the feedback!

PR #358, opened 2020-07-09
Replace DataFrame with List.

### LINKED ISSUE #357: Use a List for proxy, instead of DataFrame
Can the `DataFrame proxy` in `Las.cpp / List LAS::point_metrics(unsigned int k, double r, DataFrame data, int nalloc, SEXP call, SEXP env)` be replaced with a `List`?

I am trying to get a PR accepted into Rcpp (RcppCore/Rcpp#1099), but it breaks lidR.

The underlying cause is that previously, calling `push_back` or `push_front` on a `DataFrame` caused the `DataFrame` to revert to a list, but it should remain a `DataFrame`.
However, there are some occasions (caught by your tests, thanks!) where `proxy` becomes invalid for a `DataFrame` (to do with column lengths). It remains a valid `List`.

As far as I can tell, a `List` still serves the same role as a `DataFrame`. I have prepared a PR to implement this.

waltersom:
This fixes #357.

This is necessary, as the column length may change, resulting in an object that is not a valid `DataFrame`. The conflicts with RcppCore/Rcpp#1099 where the object will remain a `DataFrame`.

The underlying cause is that previously, calling push_back or push_front on a DataFrame caused the DataFrame to revert to a list, but it should remain a DataFrame.
However, there are some occasions (caught by your tests, thanks!) where proxy becomes invalid for a DataFrame (to do with column lengths). It remains a valid List.

For the purposes of `proxy`, a `List` is sufficient.

### DIFF
--- a/src/LAS.cpp
+++ b/src/LAS.cpp
@@ -1206,17 +1206,17 @@ List LAS::point_metrics(unsigned int k, double r, DataFrame data, int nalloc, SE
   // Retrieve the names of the attributes
   std::vector<std::string> names = as<std::vector<std::string> >(data.names());
 
-  // Need some iterator to loop on the DataFrame columns
-  Rcpp::DataFrame::iterator it1;
-  Rcpp::DataFrame::iterator it2;
+  // Need some iterator to loop on the List columns
+  Rcpp::List::iterator it1;
+  Rcpp::List::iterator it2;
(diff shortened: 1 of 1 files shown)


Jean-Romain:
This seems legit. And you are right I used `DataFrame` only as a convenience. It is perfectly valid to use a `list`.

PR #373, opened 2020-09-11
test CRS equivalence not equality

rsbivand:
In preparing for new **sp** and **rgdal** releases, I've been trying to send the SRS_string to PROJ directly, not going through GDAL to PROJ. This leads to minor differences between `sp::CRS("+init=epsg:2008")` and `sp::CRS(SRS_string="EPSG:2008")`. The WKT2 strings are equivalent but not strictly identical; the former has:
```
    CS[Cartesian,2],
        AXIS["(E)",east,
            ORDER[1],
            LENGTHUNIT["metre",1,
                ID["EPSG",9001]]],
        AXIS["(N)",north,
            ORDER[2],
            LENGTHUNIT["metre",1,
                ID["EPSG",9001]]],
    USAGE[
        SCOPE["unknown"],
        AREA["Canada - Quebec - east of 57°W"],
        BBOX[46.6,-57,53.76,-54]]] 
```
while the latter drops two `ID[]` authority declarations and adds one:
```
    CS[Cartesian,2],
        AXIS["easting (X)",east,
            ORDER[1],
            LENGTHUNIT["metre",1]],
        AXIS["northing (Y)",north,
            ORDER[2],
            LENGTHUNIT["metre",1]],
    USAGE[
        SCOPE["unknown"],
        AREA["Canada - Quebec - east of 57°W"],
        BBOX[46.6,-57,53.76,-54]],
    ID["EPSG",2008]] 
```
The differences are picked up by `expect_equal()`, but are not real differences. Using `rgdal::compare_CRS()` gives equivalence, that is the same underlying CRS but differing in unimportant textual representation. I haven't checked for **roxygen2** imports, but because **rgdal** is imported anyway, the substitution works. I also haven't checked with PROJ < 6 and GDAL < 3.

### DIFF
--- a/tests/testthat/test-projection.R
+++ b/tests/testthat/test-projection.R
@@ -8,7 +8,12 @@ test_that("Internal projection conversion works", {
   expected <- sp::CRS("+init=epsg:2008")
   expected@projargs <- sub("\\+init=epsg:\\d+\\s", "", expected@projargs)
 
-  expect_equal(lidR:::epsg2CRS(2008), expected)
+  if (rgdal::new_proj_and_gdal()) {
+    expect_true(rgdal::compare_CRS(lidR:::epsg2CRS(2008),
+      expected)["equivalent"])
+  } else {
+    expect_equal(lidR:::epsg2CRS(2008), expected)
+  }
   expect_equal(lidR:::epsg2CRS(200800), sp::CRS())
   expect_error(lidR:::epsg2CRS(200800, fail = TRUE), "Invalid epsg code")
 


Jean-Romain:
Thank you for the fix. I know I need to rework some tests relative to CRS. In which versions of `sp`, `rgdal` the former test will start to fail? To you need me to update lidR on CRAN as soon as possible?

rsbivand:
Good, Thanks! **sp** 1.4-4 will come first and should not cause trouble. **rgdal** 1.5-17 causes trouble with this, because up to then all the SRS_string= arguments are handled through GDAL, which itself calls PROJ. When working on the very difficult axis-swapping probblem, I found that PROJ seemed to give me better insight and control. So I added an SRS lookup to  PROJ directly, and made it default. The big problem is that "real" `EPSG:4326` is Latitude, Longitude by definition, but for PROJ < 6, `+init=epsg:4326` was Longitude, Latitude. We're now looking at `OGC:CRS84`, which is Longitude, Latitude by definition as a recommendation to replace `EPSG:4326`. For now both **sp** and **sf** "hope" that data is Longitude/Latitude, but as files start getting written as EPSG:4326, we'll potentially not know whether the data are standards-conformant or not. See https://lists.osgeo.org/pipermail/proj/2020-September/009802.html - this extends to EPSG:4267 too, I'm afraid.
```

**Comment by Jean-Romain** [comment]
This sounds complex... Thank you for the insight. I'll keep that in mind if somebody encounters troubles with point cloud stored as long/lat. Every single point-cloud I've seen so far was projected but it is possible to store point-cloud in LAS files as long/lat (I've heard that some files like that exist in the wild) .

I've already updated the source code to use `SRS_string` everywhere conditionally to `rgdal::new_proj_and_gdal()`.  I also made recent changes to use the WKT in `comment()` to display the CRS with  `print` (actually I performs a conversion to `sf::crs` before to print because it is more convenient). I should now update some unit tests that still use the old way `CRS("+init=epsg:xxx")`.

This is not the first time that changes in `sp` and `rgdal` trigger errors with lidR because of the slightly modified string. My goal is not to test the string but to test if a CRS was attributed properly to the LAS object according to the tools provided in `sp` and `rgdal`. Thus my tests are not properly designed. In the specific case you fixed I guess a better test should be:

```r
if (rgdal::new_proj_and_gdal())
{
   expected <- sp::CRS(SRS_string = "EPSG:2008")
   expect_equal(lidR:::epsg2CRS(2008), expected)
}
else
{
   expected <- sp::CRS("+init=epsg:2008")
   expected@projargs <- sub("\\+init=epsg:\\d+\\s", "", expected@projargs)
   expect_equal(lidR:::epsg2CRS(2008), expected)
}
```

Does this seem reliable on long term for you?

rsbivand:
It is a little hard to say, but looks reasonable, as they both follow the same path through PROJ. From sp 1.4-4, `CRS("EPSG:2008")` will also work (if the first argument does not begin  with `"+"`, it is shifted internally to SRS_string). It is hard to say not least because PROJ/GDAL are still moving, and we don't know fully what they will do.

Jean-Romain:
Ok I understand. I'll keep going with that kind of test. Internally it uses `SRS_string` so if I test against an `expected` object built with `SRS_string` it should remain valid even if the content changed. I will review all the unit tests. Thanks for your insight.

PR #380, opened 2020-10-01
Include ENL function to Update utils_metrics.R

lhambrecht:
Added function to calculate Effective Number of Layers (ENL) based on the method by Ehbrecht et al. 2016. ENL included Hill number 0D, 1D, 2D. 1D is based on the Shannon diversity index while 2D is based on the Simpson diversity index. 
Empty layers are not removed which will lead to empty cells when 1D is calculated with `grid_metrics`.

ALS and space-borne LiDAR papers often use the foliage height diversity [FHD] (based on Shannon diversity index) to assess the vertical forest structure. However, because Shannon diversity indices rely on the natural log it can not deal with empty layers. This is a really issue for high point density point clouds such as from UAV based LiDAR, dens forest due to occlusion and forest with gaps in the vertical structure.
Ehbrecht et al. proposed the ENL based on TLS scan but can be applied to ALS as well. ENL is more intuitive then FHD which is a common criticism. Furthermore, 2D ENL is based on the Simpson diversity index which is not based on the natural log and therefore can deal with empty layers. This makes it more suitable for high point density. The scalability of ENL is currently being explored but first impression seem to be promising to be able to scale ENL from TLS to ALS and even space-borne lidar.

Code needs to be reviewed but I try to based it on the entropy function.

### DIFF
--- a/R/utils_metrics.R
+++ b/R/utils_metrics.R
@@ -714,3 +714,63 @@ VCI = function(z, zmax, by = 1)
   z = z[z < zmax]
   return(entropy(z, by, zmax))
 }
+
+#' Computes the effective number of layers based on the methods Ebrecht et al. (see references)
+#' Evective Numbers of Layers (ENL) descriptor of vertical structure, and its variability across sampling points as a measure of horizontal structural variability
+#' Using diversity indices of different Hill-Numbers (^0^D, ^1^D and ^2^D, see Hill, 1973 and Jost, 2006)  
+#' $p_{i}$ is the proportion of filled voxels in the $i_{th}$ vertical layer to the sum of filled voxel in the entire volume analysed.  
+#' $N_{top}$ refers to the top stand height.  
+#' ^0^D ENL corresponds to the number of 1 m layers and thus approximates stand height ($\sum_{i=1}^n p_{i}^0$). With ^1^D ENL (exponential Shannon-Index, $\exp(-\sum_{i=1}^n p_{i} * \ln(p_{i}))$) and ^2^D ENL (inverse Simpson-Index, $1/\sum_{i=1}^n p_{i}^2$), layers are weighted stronger the more they are occupied (weighing of ^2^D ENL > that of ^1^D ENL). Weighing results in a lower number of effective layers when compared to ^0^D ENL. At a given number of layers both ^1^D and ^2^D ENL values increase with more evenly occupied layers. When considering foliage only, the natural logarithm of ^1^D ENL equals FHD based on layers of 1 m thickness.
+#' @param z vector of z coordinates
+#' @param by numeric. The thickness of the layers used (height bin)
+#' @param by numeric. Hill Number
+#' @param by numeric. Minium height cut off
+#' @references Ehbrecht, M., Schall, P., Juchheim, J., Ammer, C., Seidel, D., 2016. Effective number of layers: A new measure for quantifying three-dimensional stand structure based on sampling with terrestrial LiDAR. Forest Ecology and Management 380, 212–223. https://doi.org/10.1016/j.foreco.2016.09.003
+#' @export ENL
+ENL = function(z, by = 1, D = 2, zmin= NULL)
+{
+  # Using different Hill numbers 0-2
+  
+  
+  # Setting minium height to 0
+  if (is.null(zmin))
+    zmin = 0
+
+  # Check if D is between 0 and 2
+  if (D < 0)
+    return(NA_real_)
+
+  if (D > 2)
+    return(NA_real_)
+
+  # Define the number of x meters bins from zmin to zmax (rounded to the next integer)
+  zmax = max(z)
+  bk = seq(zmin, ceiling(zmax/by)*by, by)
+
+  # Compute the p for each bin
+  hist = findInterval(z, bk)
+  hist = fast_table(hist, length(bk) - 1)
+  hist = hist/sum(hist)
+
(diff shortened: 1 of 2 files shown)


R/utils_metrics.R
```diff
@@ -714,3 +714,63 @@ VCI = function(z, zmax, by = 1)
...
+  
+  # Setting minium height to 0
+  if (is.null(zmin))
+    zmin = 0
+
+  # Check if D is between 0 and 2
+  if (D < 0)
+    return(NA_real_)
+
+  if (D > 2)
```
Jean-Romain, line 743: The same. Actually the function should start with assertions. See`utils_assertive.R`

R/utils_metrics.R
```diff
@@ -714,3 +714,63 @@ VCI = function(z, zmax, by = 1)
...
+    return(NA_real_)
+
+  # Define the number of x meters bins from zmin to zmax (rounded to the next integer)
+  zmax = max(z)
+  bk = seq(zmin, ceiling(zmax/by)*by, by)
+
+  # Compute the p for each bin
+  hist = findInterval(z, bk)
```
Jean-Romain, line 751: Your function will fails I guess if some `z < 0` right? Unit tests will tell us.

R/utils_metrics.R
```diff
@@ -714,3 +714,63 @@ VCI = function(z, zmax, by = 1)
...
+  hist = fast_table(hist, length(bk) - 1)
+  hist = hist/sum(hist)
+
+  # Remove bins where there are no points because of log(0)
+  p    = hist[hist > 0]
+  #pref = rep(1/length(hist), length(hist))
+
+  # calculate ENL
+  if(D==2){
```
Jean-Romain, line 760: Simplify conditional statements. Use `if () { } else if ()  { } else { }` or `if () return if () return if() return`

R/utils_metrics.R
```diff
@@ -714,3 +714,63 @@ VCI = function(z, zmax, by = 1)
...
+    enl <- 1 / sum(p^2) #2D ENL formula
+  }else{
+    if(D==1){
+      enl <- exp(-sum(p*log(p))) #1D ENL formula
+    }else{
+      if(D==0){
+      enl <- sum(p^0) #0D ENL formula
```
Jean-Romain, line 767: Hmm this is equivalent to `length(p)` right?

R/utils_metrics.R
```diff
@@ -714,3 +714,63 @@ VCI = function(z, zmax, by = 1)
...
+      }else{
+      stop("D must be 0,1 or 2.") #Error
```
Jean-Romain, line 769: Should be tested *before* to perform any costly computation and throw an error

R/utils_metrics.R
```diff
@@ -714,3 +714,63 @@ VCI = function(z, zmax, by = 1)
...
+#' ^0^D ENL corresponds to the number of 1 m layers and thus approximates stand height ($\sum_{i=1}^n p_{i}^0$). With ^1^D ENL (exponential Shannon-Index, $\exp(-\sum_{i=1}^n p_{i} * \ln(p_{i}))$) and ^2^D ENL (inverse Simpson-Index, $1/\sum_{i=1}^n p_{i}^2$), layers are weighted stronger the more they are occupied (weighing of ^2^D ENL > that of ^1^D ENL). Weighing results in a lower number of effective layers when compared to ^0^D ENL. At a given number of layers both ^1^D and ^2^D ENL values increase with more evenly occupied layers. When considering foliage only, the natural logarithm of ^1^D ENL equals FHD based on layers of 1 m thickness.
+#' @param z vector of z coordinates
+#' @param by numeric. The thickness of the layers used (height bin)
+#' @param by numeric. Hill Number
+#' @param by numeric. Minium height cut off
+#' @references Ehbrecht, M., Schall, P., Juchheim, J., Ammer, C., Seidel, D., 2016. Effective number of layers: A new measure for quantifying three-dimensional stand structure based on sampling with terrestrial LiDAR. Forest Ecology and Management 380, 212–223. https://doi.org/10.1016/j.foreco.2016.09.003
+#' @export ENL
```
Jean-Romain, line 729: Missing `@examples`

Jean-Romain (changes requested):
Thank you for integrating your work in `lidR`. Sadly, as is, I cannot accept the PR. I put many comments + here a list of thing do to:

- [ ] Generate the `Rd` file of the documentation to check that you documentation is valid (it is not). And include the Rd file in the PR
- [ ] Write unit tests in `test/testthat/test-metrics.R`. Do not take example on the existing tests in this file they are weak. Your test must ensure that your function fails properly when needed, returns the good value in all cases, test limit cases with unexpected data and so on...
- [ ] Run `R CMD check`.  As is your PR fails CRAN checks
- [ ] Add examples in the doc.
- [ ] Update contributors in `DESCRIPTION` files.

Also I only looked at section 2.2.2.1 quickly in the paper. I see:

> where p_i is the proportion of filled voxels in the ith vertical layer to the sum of filled voxel in the entire volume analyzed.

Which is definitively not what is computed here. Please explain me. 

Thanks.

lhambrecht:
Thanks for the feedback. This is much appreciated since this is my first time contributing on GitHub.
I will try to address these issue the best I can.
BTW, congratulations on the paper!

Jean-Romain:
This PR has been opened 1 month ago. Do you plan to finalize the PR or can I close it?

PR #611, opened 2022-08-04
Don't update display after every sphere.

dmurdoch:
Because spheres are added one at a time, the full display will be redrawn after each one, and that is very slow.  This will speed it up quite a bit.

### DIFF
--- a/R/algorithm-itd.R
+++ b/R/algorithm-itd.R
@@ -195,8 +195,18 @@ manual = function(detected = NULL, radius = 0.5, color = "red", button = "middle
     plot.LAS(las, ..., clear_artifacts = FALSE)
 
     id = numeric(nrow(apice))
+
+    # It's very slow to redraw after every sphere, so
+    # turn off updates for a bit
+    saveSkip = rgl::par3d(skipRedraw = TRUE)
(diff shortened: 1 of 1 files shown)


Jean-Romain:
Thank you for this. I discovered this option yesterday after the discussion on stack-overflow (it was you btw) and this changed the way I understand `rgl`. This change was in my plan but thank you for the commit.

Following the discussion started [here](https://github.com/dmurdoch/rgl/issues/260) and [here](https://stackoverflow.com/questions/73225889/r-rgl-lidr-slow-rendering-on-windows-11-64-bit) about rendering very huge point cloud like `1e8+` points I'm currently thinking about a spatially indexed rendering that only displays `n` points. The idea is to group the points into chunks at different level of resolution and display the point cloud by chunks. When the user zoom-in or out we can recompute which chunks are visible or not and add new chunks if necessary and remove chunks if no longer visible. The goal being to maintain a constant amount of points (e.g `1e6`) and increase the local density gradually when the user is zooming. With `pop3d` and `saveSkip = rgl::par3d(skipRedraw = TRUE)` I think it is doable to design something smooth.

Issue #726, opened 2023-11-07
Character palette causes error


dmurdoch:
This line: 
https://github.com/r-lidar/lidR/blob/dd16f690565a370e30eb369ad5c0bee11f8f81fb/R/plot.R#L290C35-L290C35 
causes an error if the palette is specified as a vector of colors.

A simple fix is to change it to 

      autocol <- is.character(pal) && length(pal) == 1 && pal == "auto"

Here is code to illustrate:

     LASfile <- system.file("extdata", "MixedConifer.laz", package="lidR")
     las <- readLAS(LASfile)
     lidR::plot(las,size=5,pal=grDevices::terrain.colors(30),axis=TRUE)

This was reported [here](https://stackoverflow.com/q/77434504).

PR #727, opened 2023-11-07
If a palette is character, check length before comparing to fixed string

dmurdoch:
Fixes issue #726


Jean-Romain:
Thank you for the fix

PR #792, opened 2024-12-19
Improve epsg() warning message when WKT present

ppoyk:
EPSG code in the LAS header might be correctly populated for backwards compatability, despite also having a defined WKT as per LAS-format 1.4 requirements.
It is best for the method to return the EPSG contents but with a warning. Previous warning message ("0 returned") is untrue if EPSG is populated.


Jean-Romain:
That makes sense thank you.

PR #841, opened 2025-12-26
Accommodate R-devel changes

eddelbuettel:
Salut @Jean-Romain,

As discussed the other day over in lasR repo in this org, Luke Tierney is tightening the R API.  In R's move away from ATTRIB() we changed how we compute the number of rows in a data.frame (it is easier and better now) but also changed to returning a larger R_xlen_t.  It turns out your file src/knn.cpp is the only one using this in a type-checking context such as the `std::min()` so it needs a cast.   

When looking at this I noticed CRAN is also nagging you about ATTRIB so I injected a C++ alternative with a 'visitor' passed to the (allowed, added in R 4.6.0 aka R-devel) R_mapAttrib.  With both these changes, the packages passes all checks under R-devel.

There are two small things remaining.  Kurt added more URL checks, so your manual pages complain:

```
* checking CRAN incoming feasibility ... [2s/12s] NOTE          
Maintainer: ‘Jean-Romain Roussel <info@r-lidar.com>’
                                                    
Found the following (possibly) invalid URLs:   
  URL: https://www.asprs.org/divisions-committees/lidar-division/laser-las-file-format-exchange-activities
    From: man/add_attribute.Rd          
          man/classify.Rd                                                                                
          man/las_utilities.Rd
          man/LAS-class.Rd   
          man/LASheader-class.Rd
          man/rasterize.Rd
    Status: 404
    Message: Not Found
* checking package namespace information ... OK
```

That should be easy to fix.  More annoyingly, Luke has one more for you on 

```
* checking compiled code ... NOTE
File ‘lidR/libs/lidR.so’:
  Found non-API call to R: ‘SETLENGTH’

Compiled code should not call non-API entry points in R.

See ‘Writing portable packages’ in the ‘Writing R Extensions’ manual,
and section ‘Moving into C API compliance’ for issues with the use of
non-API entry points.
* checking installed files from ‘inst/doc’ ... OK
```

There is an entire new section [Resizing Vectors](https://cran.r-project.org/doc/manuals/r-devel/R-exts.html#Resizing-vectors) in WRE that I can point you too, but I didn't want to make invasive changes. It seems like if you declare a vector 'resizable' on allocation you are allowed to shrink it.

Hope this helps, and once again thanks for considering these changes.  Happy Holidays, and a Happy 2026!

### DIFF
--- a/src/altrep.cpp
+++ b/src/altrep.cpp
@@ -16,6 +16,7 @@ bool is_materialized(SEXP x)
 // [[Rcpp::export]]
 SEXP altrep_full_class(SEXP x) {
   if (!ALTREP(x)) return R_NilValue;
+#if R_VERSION >= R_Version(4, 6, 0)
   std::vector<std::string> v;
   auto visitor = [](SEXP name, SEXP attr, void* data) -> SEXP {
     std::vector<std::string>* ptr = static_cast<std::vector<std::string>*>(data);
@@ -25,4 +26,7 @@ SEXP altrep_full_class(SEXP x) {
   };
   R_mapAttrib(ALTREP_CLASS(x), visitor, static_cast<void*>(&v));
   return Rcpp::wrap(v);
+#else
+  return ATTRIB(ALTREP_CLASS(x));
+#fi
 }
--- a/src/altrep.cpp
+++ b/src/altrep.cpp
@@ -15,7 +15,14 @@ bool is_materialized(SEXP x)
 
 // [[Rcpp::export]]
 SEXP altrep_full_class(SEXP x) {
-  if (ALTREP(x)) return ATTRIB(ALTREP_CLASS(x));
-  return R_NilValue;
+  if (!ALTREP(x)) return R_NilValue;
+  std::vector<std::string> v;
+  auto visitor = [](SEXP name, SEXP attr, void* data) -> SEXP {
+    std::vector<std::string>* ptr = static_cast<std::vector<std::string>*>(data);
+    std::string s{CHAR(Rf_asChar(attr))};
+    ptr->push_back(s);
+    return NULL;
+  };
+  R_mapAttrib(ALTREP_CLASS(x), visitor, static_cast<void*>(&v));
+  return Rcpp::wrap(v);
 }
-
--- a/src/knn.cpp
+++ b/src/knn.cpp
@@ -93,7 +93,7 @@ List cpp_knnx(S4 data, S4 query, int k, int ncpu)
   NumericVector Z = tmp["Z"];
 
   int npoints = X.size();
-  k = std::min(k, temp.nrow());
+  k = std::min(static_cast<R_xlen_t>(k), temp.nrow());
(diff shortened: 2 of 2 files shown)


Jean-Romain:
Thank you for the PR, it’s really appreciated. That will save me hours.

Regarding the URL, there’s nothing I can do. The issue is known and the URLs are correct; the server is simply rejecting CRAN (and likely other bot-like requests).

As for `SETLENGTH`, this is not an issue. It appears because you ran `R CMD check`  as-is. By default, the code is compiled with `-DWITHSETLENGTH` on GitHub and r-universe to preserve features that have been removed on CRAN. I’ve been submitting the package to CRAN without this feature for the past two years, since resizable vectors trigger issues. This part of the code is therefore not included on CRAN. I'm waiting better programmers to find a workaround (e.g. `data.table` team). The link you shared look like the first workable workaround I've seen in two years. Thank you.

eddelbuettel:
Makes sense for SETLENGTH.  I have not had a need to reduce vector size (apart from the obious 'copy-and-destroy-old-one') but like you I am aware that the `data.table` team insists it is key (because IIRC they start overallocating to speculatively save some cycles...).  

Glad the PR worked for you.  I once again learned something new about ALTREP.

eddelbuettel:
Do you mean with Rcpp?  Yes.  It required one other package to change, it is also now in incoming.  What is in the Rcpp repo will be 1.1.1 once @kurthornik and I dot all i-s and cross all t-s on URLs.  Code will not change -- Rcpp 1.1.0.14 as committed will be the same as 1.1.1.  A of purely cosmetic README and vignette and .bib changes in the air.

Thanks for the prompt update of lidR -- much appreciated.

Jean-Romain:
sent on CRAN. Thanks

Jean-Romain:
So it did not pass CRAN pretest on windows

```
* checking compiled code ... NOTE
File 'lidR/libs/x64/lidR.dll':
  Found non-API call to R: 'ATTRIB'

Compiled code should not call non-API entry points in R.
```

despite is passed on win-builder

Jean-Romain:
Ok I understand. So I replied to the CRAN. I hope it gonna pass with manual validation. I know that a NOTE is not a WARNING but usually a rejection is a rejection even with a NOTE. But the explantions sound strong :smile:

eddelbuettel:
Yes. I wrote a little helper package `ciw` (== CRAN Incoming Watcher) so that `ciw::ciw()` shows (quickly, easily) what is in the queue and `lidR` got moved to section 'archive/' which is puzzling.  I hope this gets straightened out -- it also happens to some packages of mine (e.g. Rcpp has one old grandfathered exception but it still comes up each time -- sigh).  That is to say:  email to them should help explain it.

kurthornik:
>>>>> Kurt Hornik writes:

And meanwhile on CRAN ...

-k

eddelbuettel:
All good -- thanks so much for the prompt update!

